PR #3, opened 2022-11-21
[Fetch] Orderbook Rewards

bh2smith:
We migrate a query we have seen before in `solver-rewards` (namely the one querying for orderbook rewards on a block range). This block range will be specified as:

block_from = the latest synced block (coming in a follow up PR)
block_to = the latest synced block from the two orderbook databases

block_to is implemented here by querying the two databases for the smallest block with `tx_from = Null` taking the minimum of these two and subtracting 65 (for REORG protection)

### Follow Up

https://github.com/cowprotocol/dune-sync/issues/6

## Test Plan

Tests are introduced for both database queries (however they require database credentials).

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -20,7 +20,7 @@ jobs:
         run: black --check ./
       - name: Type Check (mypy)
         run: mypy src --strict
-  unit-tests:
+  tests:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2


gentrexha:
@bh2smith Some of the unit tests are failing is this still to be reviewed?

bh2smith:
> @bh2smith Some of the unit tests are failing is this still to be reviewed?

I am just getting the DB secrets configured here. Just a moment.

src/environment.py
```diff
@@ -0,0 +1,5 @@
+"""Project Global Constants."""
```
gentrexha, line 1: In my opinion these could belong to the `utils.py` file.

tests/test_fetch_orderbook.py
```diff
@@ -0,0 +1,45 @@
+import json
+import unittest
+
+import pandas as pd
+
+from src.fetch.orderbook import OrderbookFetcher
+from src.models.block_range import BlockRange
+
+
+class MyTestCase(unittest.TestCase):
```
gentrexha, line 10: In my opinion we should split these tests into two separate categories: `unit` and `integration`. What's your take on that?

gentrexha (approved):
Good work Ben!

Left some minor discussion comments for moving things around. Other than that LGTM!

src/environment.py
bh2smith: Perhaps, at this moment that might make sense, but if either of these two things starts to go, the purpose for the single files might diverge and they would have to be split up again.


In fact, we already have PROJECT_ROOT being used in more than one file `main.py` and `utils.py` which already indicates the need for two files.

We could move the QUERY_PATH into utils since this is the only place its used, but then people would have to go to two different files to see the global configuration....

I am thinking that it is fine the way it is.

tests/test_fetch_orderbook.py
bh2smith: Don't worry I will fix this!

PR #9, opened 2022-12-02
Configuration Final Touches

bh2smith:
Preparing for deployment on kubernetes we fix a few things:

- using an absolute VOLUME_PATH (instead of relative to the project root)
- configure log level via runtime arguments
- use AWS access keys instead of PROFILE (so we don't have to rely on a credentials existing within the file system).

## Test Plan

Integration tests (however they all require A LOT of credentials). You could run the main branch in `dry-run` mode with empty aws credentials (however, there was a subtle bug discovered here while testing that is resolved in a follow up PR #10). 

```sh
python -m src.main --sync-table order_rewards --dry-run True
python -m src.main --sync-table app_data --dry-run True  <--! Running This from scratch takes over an hour!
```

### DIFF
--- a/src/logger.py
+++ b/src/logger.py
@@ -9,7 +9,7 @@ from src.environment import LOG_LEVEL
 def set_log(name: str) -> Logger:
     """
     Instantiates and returns a logger with the name defined by the caller.
-    Used by any file which wants to inherit the global log level configuration 
+    Used by any file which wants to inherit the global log level configuration
     and assigns the desired naming convention (usually `__name__`)
     """
 
--- a/src/logger.py
+++ b/src/logger.py
@@ -1,0 +1,19 @@
+"""Easy universal log configuration """
+import logging.config
+import sys
+from logging import Logger
+
+from src.environment import LOG_LEVEL
+
+
+def set_log(name: str) -> Logger:
+    """Removes redundancy when setting log in each file"""
+
+    log = logging.getLogger(name)
+    try:
+        log.setLevel(level=LOG_LEVEL)
+    except ValueError:
+        logging.error(f"Invalid log level: {LOG_LEVEL}")
+        sys.exit(1)
+
+    return log
--- a/src/logger.py
+++ b/src/logger.py
@@ -7,7 +7,11 @@ from src.environment import LOG_LEVEL
 
 
 def set_log(name: str) -> Logger:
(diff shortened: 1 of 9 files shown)


src/sync/order_rewards.py
```diff
@@ -1,19 +1,17 @@
...
 from pandas import DataFrame
 
 from src.fetch.orderbook import OrderbookFetcher
+from src.logger import set_log
 from src.models.block_range import BlockRange
 from src.sync.common import last_sync_block, aws_login_and_upload
 from src.sync.config import SyncConfig
 
-log = logging.getLogger(__name__)
-logging.basicConfig(format="%(asctime)s %(levelname)s %(name)s %(message)s")
```
MartinquaXD, line 15: Looks like this formatting was used a lot and I don't see any place where `logging` initializes this format. Is that intended?

MartinquaXD (commented):
Looks reasonable to me

src/sync/order_rewards.py
bh2smith: Turns out this formatting is the default and so I don't see any need to set it (sorry I didn't mention it).

src/logger.py
```diff
@@ -0,0 +1,19 @@
+"""Easy universal log configuration """
+import logging.config
+import sys
+from logging import Logger
+
+from src.environment import LOG_LEVEL
+
+
+def set_log(name: str) -> Logger:
```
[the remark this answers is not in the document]
bh2smith: Yea, possibly more appropriate but soooo_much_longer...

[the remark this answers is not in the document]
bh2smith: Will do.

src/logger.py
```diff
@@ -0,0 +1,19 @@
...
+    """Removes redundancy when setting log in each file"""
```
bh2smith, line 10:
```suggestion
    """
    Instantiates and returns a logger with the name defined by the caller.
    Used by any file which wants to inherit the global log level configuration 
    and assigns the desired naming convention (usually `__name__`)
    """
```

bh2smith: How does this look @MartinquaXD ?

MartinquaXD: Nice 👍

MartinquaXD (approved):

PR #10, opened 2022-12-02
[Fix] DryRun wasn'

bh2smith:


src/main.py
```diff
@@ -59,10 +59,10 @@ def __init__(self) -> None:
...
     load_dotenv()
     aws = AWSData(
-        internal_role=os.environ["AWS_INTERNAL_ROLE"],
-        external_role=os.environ["AWS_EXTERNAL_ROLE"],
-        external_id=os.environ["AWS_EXTERNAL_ID"],
-        bucket=os.environ["AWS_BUCKET"],
+        internal_role=os.environ.get("AWS_INTERNAL_ROLE", "AWS_INTERNAL_ROLE"),
+        external_role=os.environ.get("AWS_EXTERNAL_ROLE", "AWS_EXTERNAL_ROLE"),
+        external_id=os.environ.get("AWS_EXTERNAL_ID", "AWS_EXTERNAL_ID"),
+        bucket=os.environ.get("AWS_BUCKET", "AWS_BUCKET"),
```
MartinquaXD, line 65: What is the purpose of using fallback values here? They don't look like they would be useful in a program.
Should this maybe return an error if the keys don't exist?

src/sync/record_handler.py
```diff
@@ -0,0 +1,78 @@
...
+                log.info(
+                    f"{self.name} sync for block range {self.block_range} complete: "
+                    f"synced {self._num_records()} records"
+                )
+        else:
+            log.info(
+                f"No new {self.name} for block range {self.block_range}: no sync necessary"
+            )
+
+        self._write_sync_data()
```
MartinquaXD, line 78: Calling `_write_sync_data()` here looks strange. Is the intention here to let the base class call the deriving class's implementation of `_write_sync_data()` (because the base class's implementation is a no-op)?
Usually it only works the other way around (the derived class calls the implementation of the base class) and I would assume this is also how `abc` implements it.

src/main.py
bh2smith: Problem is that they are read at the entry point of the program. And dry run mode doesn't actually require them. We wouldn't want to fail with a key error when we don't actually need them. 

I could move the reading of these environment variables to where they are needed, but then the program will run without knowing it's doesn't have what it needs. 

Maybe I will fall back to key error but put an if block here that doesn't try to read them in dry run mode. 

How does that sound? Other suggestions?

src/sync/record_handler.py
bh2smith: Maybe I just don't understand the abstraction. I would assume it calls the method. I'm pretty sure that's what it does since the abstraction forces the implementation to be there.

MartinquaXD: Most of the time the point of abstract classes is to have some (optional) shared code together with some sort of interface. Then multiple classes can implement the same interface (by overriding the abstract methods) while being able to share the (optional) shared code of the abstract base class.
This however does not mean that the base class is able to call the code from classes that got derived from itself. Usually classes can only access their own code and the code from its parent but not from its children so to speak.

MartinquaXD: From the `abc` docs:
```
Note
Unlike Java abstract methods, these abstract methods may have an implementation. This implementation can be called via the super() mechanism from the class that overrides it. This could be useful as an end-point for a super-call in a framework that uses cooperative multiple-inheritance.
```
And docs for [`super()`](https://docs.python.org/3/library/functions.html#super).

src/main.py
MartinquaXD: >Maybe I will fall back to key error but put an if block here that doesn't try to read them in dry run mode.

If you already know at that point that you are running in `dry-run` mode this approach makes the most sense to me.

src/sync/record_handler.py
bh2smith: So this is all fine and dandy, but I don't know what other pattern I am supposed to use instead. This code is doing what I expect -- do you happen to know what is the correct abstraction to be using for this purpose?

MartinquaXD: What you could do is:
```
class RecordHandler {
def __init__(
        self,
        block_range: BlockRange,
        config: SyncConfig,
    ):
        self.config = config
        self.block_range = block_range

        self.name = config.table_name
        self.file_path = config.volume_path / self.name
        self.content_filename = f"cow_{block_range.block_to}.json"
        
    @abstractmethod
    def num_records(self) -> int:
        """Returns number of records to handle"""

    @abstractmethod
    def write_found_content(self) -> None:
        """Writes content to disk"""

    @abstractmethod
    def write_sync_data(self) -> None:
        """Records last synced content file"""
        
    def upload_to_aws(self, dry_run:bool) -> None:
               if dry_run:
                log.info(
                    "DRY-RUN-ENABLED: New records written to volume, but not posted to AWS."
                )
            else:
                aws_login_and_upload(
                    config=self.config,
                    path=self.file_path,
                    filename=self.content_filename,
                )
                log.info(
                    f"{self.name} sync for block range {self.block_range} complete: "
                    f"synced {self._num_records()} records"
                )
}
```

Then `AppDataHandler` and `OrderookDataHandler` could implement the abstract functions of `RecordHandler`.

And then you could have a `UploadHandler` which takes a `RecordHandler` which is now an interface which is implemented by `AppDataHandler` and `OrderookDataHandler` so you should be able to pass in either of those types into `UploadHandler`.
And finally `UploadHandler` manages the upload like this:
```
class UploadHandler
def __init__(handler: RecordHandler):
    self.handler = handler

def write_and_upload_content(self, dry_run: bool) -> None:
        """
        - Writes self.order_rewards to persistent volume,
        - attempts to upload to AWS and
        - records last sync block on volume.
        When dryrun flag is enabled, does not upload to IPFS.
        """
        super.write_found_content()
        if handler.num_records() > 0:
            log.info(
                f"attempting to post {handler.num_records()} new {self.name} "
                f"records for block range {self.block_range}"
            )
            handler.upload_to_aws()
        else:
            log.info(
                f"No new {self.name} for block range {self.block_range}: no sync necessary"
            )

        handler.write_sync_data()
```

This allows `AppDataHandler` and `OrderookDataHandler` to share some common code while also implementing some custom code and `UploadHandler` can just delegate to the concrete implementations.

MartinquaXD: You could also make `RecordHandler` completely abstract and move `upload_to_aws` into `UploadHandler`. This is probably more correct.

bh2smith: Thanks! I will do this!

bh2smith: The only issue I see with this approach is the fact that I would have to make all the private methods public. They are meant to be private so that end users can not call them in the incorrect order. I am not so sure I want to do this...

bh2smith: Okie - Thanks for this! Its much better now, although I think there is still room for some further generalizations I will save for later.

src/main.py
bh2smith: Done!

src/sync/record_handler.py
MartinquaXD: Cool. Now it looks like it would call the expected code. 👍

MartinquaXD (approved):

PR #15, opened 2022-12-06
IPFS Fetch Content Async

bh2smith:
This makes the program faster (I hope), but also required quite a few changes. It did lead me to discover that we weren't very string with our handling of app data content (we were manipulating dictionaries). We also introduce a data structure for the Found and not found app data types.

### DIFF
--- a/src/fetch/ipfs.py
+++ b/src/fetch/ipfs.py
@@ -97,9 +97,6 @@ class Cid:
                             break
                     except asyncio.TimeoutError:
                         attempts += 1
-                    except Exception as err:
-                        log.error(f"unexpected error {err} -- requires investigation, retrying anyway")
-                        attempts += 1
 
                 if not content:
                     total_attempts = previous_attempts + max_retries
--- a/src/fetch/ipfs.py
+++ b/src/fetch/ipfs.py
@@ -55,6 +55,9 @@ class Cid:
                 return response.json()
             except requests.exceptions.ReadTimeout:
                 attempts += 1
+            except requests.exceptions.JSONDecodeError as err:
+                attempts += 1
+                log.warning(f"unexpected error {err} retrying...")
         return None
 
     @classmethod
@@ -93,6 +96,9 @@ class Cid:
                             )
                             break
                     except asyncio.TimeoutError:
+                        attempts += 1
+                    except Exception as err:
+                        log.error(f"unexpected error {err} -- requires investigation, retrying anyway")
                         attempts += 1
 
                 if not content:
--- a/src/fetch/ipfs.py
+++ b/src/fetch/ipfs.py
@@ -80,7 +80,9 @@ class Cid:
                                     f"Found previously missing content hash {app_hash} at CID {cid}"
                                 )
                             else:
-                                log.debug(f"Found content for {app_hash} at CID {cid}")
+                                log.debug(
+                                    f"Found content for {app_hash} at CID {cid} ({attempts + 1} trys)"
+                                )
                             found.append(
                                 FoundContent(
                                     app_hash=app_hash,
--- a/src/fetch/ipfs.py
+++ b/src/fetch/ipfs.py
@@ -1,10 +1,17 @@
 """IPFS CID (de)serialization"""
 from __future__ import annotations
 
+import asyncio
 from typing import Any, Optional
 
+import aiohttp
 import requests
 from multiformats_cid.cid import from_bytes
+
+from src.logger import set_log
+from src.models.app_data_content import FoundContent, NotFoundContent
+
+log = set_log(__name__)
 
 
 class Cid:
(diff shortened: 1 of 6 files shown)


src/fetch/ipfs.py
```diff
@@ -49,3 +56,55 @@ def get_content(self, max_retries: int = 3) -> Optional[Any]:
             except requests.exceptions.ReadTimeout:
                 attempts += 1
         return None
+
+    @classmethod
+    async def fetch_many(
+        cls, missing_rows: list[dict[str, str]], max_retries: int = 3
+    ) -> tuple[list[FoundContent], list[NotFoundContent]]:
```
MartinquaXD, line 63: My assumption is that you wanted to speed up the program by fetching multiple rows concurrently. I have no experience with asynchronous operations in python but I'm pretty sure the equivalent code in rust wouldn't do that.

Looks like in order to have true concurrency you would have to:
1. collect the async requests in a list
2. await all of them concurrently with `await asyncio.gather(requests)`

[stackoverflow reference](https://stackoverflow.com/a/47172471)

src/models/app_data_content.py
```diff
@@ -0,0 +1,59 @@
...
+            first_seen_block=int(row["first_seen_block"]),
+            attempts=int(row["attempts"]),
+        )
+
+    def as_dune_record(self) -> dict[str, Any]:
+        """Converts to DuneRecord type"""
+        return {
+            "app_hash": self.app_hash,
+            "first_seen_block": self.first_seen_block,
+            "attempts": self.attempts,
```
MartinquaXD, line 58: Does it make sense to store the `attempts` of a non found record on dune? Is that relevant or useful for anybody?

src/sync/app_data.py
```diff
@@ -55,73 +55,44 @@ def num_records(self) -> int:
...
-                self._found.append(
-                    {
-                        "app_hash": app_hash,
-                        "first_seen_block": row["first_seen_block"],
-                        "content": app_data,
-                    }
-                )
-            elif app_data is None and attempts > give_up_threshold:
+    async def _handle_missing_records(
+        self, max_retries: int, give_up_threshold: int
```
MartinquaXD, line 63: Is the idea here for example to have a `give_up_threshold` of 9 and `max_retries` of 3 in order to try fetching a file at most 3 times with 3 retries per iteration? That would seem needlessly complex.

Instead you could write `fetch_many()` to only do one request per AppData per call (and update `attempts` accordingly). And then `_handle_missing_records()` could simply drop all AppDatas from the `missing` list that exceeded the `max_retries`. That way you should have minimum latency while only needing one `retry` parameter.

src/models/app_data_content.py
bh2smith: No this is not something we need to make available in dune. Our sync is append only and if we don't have the content then we will not post it until we are "sure" there is no content available.

src/fetch/ipfs.py
bh2smith: So I definitely agree that this does not look like it does what we would expect, but I was following [this article](https://www.twilio.com/blog/asynchronous-http-requests-in-python-with-aiohttp) and I believe the async aspect of this is encapsulated entirely within this clause

```
async with aiohttp.ClientSession() as session:
```

So that we can functionally manipulate the data we get within our "gather" statement.

I have tested this locally and we do indeed experience a substantial performance improvement.

src/fetch/ipfs.py
```diff
@@ -49,3 +56,55 @@ def get_content(self, max_retries: int = 3) -> Optional[Any]:
...
+        """Async AppData Fetching"""
+        found, not_found = [], []
+        async with aiohttp.ClientSession() as session:
```
bh2smith, line 66: This thing @MartinquaXD - I believe to be equivalent to `asycio.gather()` as is an alternate (less easy to work with variant).

src/fetch/ipfs.py
MartinquaXD: I skimmed the article and section `Utilizing asyncio for improved performance` sounds like what you are describing here.
>So 8 seconds compared to 29 seconds is a huge jump in performance, but we can do even better using the tools that asyncio provides. In the original example, we are using await after each individual HTTP request, which isn't quite ideal. It's still faster than the requests example because we are running everything in coroutines, but we can instead run all of these requests "concurrently" as asyncio tasks and then check the results at the end, using asyncio.ensure_future and asyncio.gather.

src/sync/app_data.py
bh2smith: Yes! I totally agree that this very strange approach (as you have accurately assumed it works) and I have a hunch that it does not need to be so complex. This logic is currently mimicking how we fetch app data in the [legacy setup](https://github.com/cowprotocol/dune-bridge/blob/c07bc1ff173c58e63d6ad8f89a936601c76a1eec/src/referral_maintenance.rs)

I believe the idea is the following: It is well known that the first attempt with such a low timeout=1s does need to be retried more than once at a time. However, sometimes the issue can be with the IPFS node provider being unavailable, so we don't entirely give up on a single run of the cron job, but try again on the next run loop (to a maximum). 

I would however, need to test the theory (could probably do it here) and collect some metrics on how many retries it actually takes to get the results -- or if on average recovered app content is actually recovered within the first attempt (I do not belive this to be true, because I have experienced the scenario in this program where some app data is not recovered in a single run - with 3 retrys, but is found on the next run). 

I understand that this is a bit ugly/confusing but I do think we should keep it like this. I could however collect some data to see how many retries are actually needed on average to get the app data on the first run. What do you think?

MartinquaXD: Good to know. Would be interesting to see how often it happens that an `AppData` can be fetched at the 2nd or 3rd try within the same chronjob instance. If this number is very low one could consider to always do a single try per chronjob only to make the logic a bit less complex.

bh2smith: I will do some naive benchmarking here in this unit test (by logging a warning every time attempts > 1) for a large set. Lets see how it plays out.

src/fetch/ipfs.py
bh2smith: So would you say that this comment is "resolved"?

bh2smith: Actually, I am not sure this even matters. The only time we need this performance boost is on the first run before any of the data has been synced. There are ~2300 appData hashes and they are all picked up on the very first run of this service. once this table has been populated with these records, there are only ever very sparsely any new one and the service doesn't really need to be async. It could happen that we need to drop the table and rebuild although this is not expected to be very often. I think its nice to have a slight performance boost, but since it only ever matters for a "genesis" run of the service I am not sure it needs to be over done. I also fine the gather approach to be more complicated to implement with all this internal data manipulation that needs to be made.

MartinquaXD: In that case this comment can be considered "resolved".

src/sync/app_data.py
bh2smith: On a single run with 84 records we got: `attempts => count`

```
{
	1: 72,
	2: 10,
	4: 1,
        5: 1
}
```

bh2smith: Another run: `{1: 77, 2: 5, 4: 1, 5: 1}`
Another run: `{2: 82, 4: 2}`

I think the 3 retries per run seems to be pretty good at capturing this, but please let me know what you think.

MartinquaXD: Yeah, seems alright. Probably better to not introduce unexpected problems by changing the code unnecessarily.

bh2smith: So in that case I believe we can call this resolved.

bh2smith:
I believe all issues raised by @MartinquaXD have been resolved. How do you feel about merging this?

MartinquaXD (approved):

PR #18, opened 2022-12-08
[Schema] Order Rewards to use JSON field for variable data.

bh2smith:
The schema proposed here is `[order_uid: str | tx_hash: str | solver: str | data ]` where (currently) data takes the following form:
```
Data {
   surplus_fee: str (we can make make this nullable, or use zero in place of null)
   amount: float
   safe_liquidity: boolean (nullable)
}
```
Examples, of this are provided in the unit test.

### Additional Notes
1. Technically the `solver` column isn't necessary since it can be obtained from the `tx_hash`. Let me know if we think it should be removed.

2. I spoke with Dune today about string types in spark SQL and their answer was as I expected:

> We use decimal(38,0) or just strings for large integers in spark. DuneSQL will provide native support for large integers but it’s not directly available in spark sql

Note that, we will use strings for the `surplus_fee` and cast to Decimal in the spell that parses this content.

## Test Plan

with all the appropriate credentials you can run:

```
python -m src.main --sync-table order_rewards --dry-run True
```

Which will process orders and write locally.
A sample output with two rows would look like this:

```
{"order_uid": "0xa386250eda29e80d804b228908968865ab31fa8023abc5bbbf532ba965ecaa6f844eb02130e1849bf94e15ef08cdf070fbf1302b639270eb", "tx_hash": "0x2a6ccd340d7fa12d483851d0d7fa5ae6627ec852a2e3c638bec3deac1c1837be", "solver": "0x97ec0a17432d71a3234ef7173c6b48a2c0940896", "data": {"surplus_fee": "0", "amount": 40.21522326123023, "safe_liquidity": null}}
{"order_uid": "0xb0677cedb8279bfde76e6c15cc264b61386d2e2b4ddeb76d59cea0a635c59eb0b2515a7221b2654f9faae0e4ed1d0e49aa7b85dd63927116", "tx_hash": "0x7e482a5b04dc78d64a36588e90d549ecfb2d5d9f90d261848bbfc3269758cda5", "solver": "0xc9ec550bea1c64d779124b23a26292cc223327b6", "data": {"surplus_fee": "0", "amount": 38.97091536669401, "safe_liquidity": null}}
```

### DIFF
--- a/src/models/order_rewards_schema.py
+++ b/src/models/order_rewards_schema.py
@@ -1,0 +1,69 @@
+"""Model for Order Rewards Data"""
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+from pandas import DataFrame
+
+
+@dataclass
+class OrderRewards:
+    """
+    This class provides a transformation interface for the Dataframe we fetch from the orderbook
+    Within are several methods for converting from the Dataframe to list[DuneRecords]
+    """
+
+    order_uid: str
+    tx_hash: str
+    solver: str
+    data: dict[str, Any]
+
+    @classmethod
+    def from_pdf(cls, rewards_df: DataFrame) -> list[OrderRewards]:
+        """Multi Record Constructor from Pandas Dataframe"""
+        return [
+            cls(
+                order_uid=row["order_uid"],
+                tx_hash=row["tx_hash"],
+                solver=row["solver"],
+                data={
+                    # This seems like a hack. I need to experiment with Decimal type
+                    # CAST("18446744073709001111111111" AS DECIMAL(38,0));
+                    # but this can be done on the spellbook side.
+                    "surplus_fee": str(row["surplus_fee"]),
+                    "amount": float(row["amount"]),
+                    "safe_liquidity": row["safe_liquidity"],
+                },
+            )
+            for row in rewards_df.to_dict(orient="records")
+        ]
+
+    def as_dune_record(self) -> dict[str, Any]:
+        """Convert to Dune Record"""
+        return {
+            "order_uid": self.order_uid,
+            "tx_hash": self.tx_hash,
+            "solver": self.solver,
+            "data": self.data,
+        }
+
+    @classmethod
+    def from_pdf_to_dune_records(cls, rewards_df: DataFrame) -> list[dict[str, Any]]:
+        """
+        The above two methods require two iterations of the entire record set
+        while this one combines both and requires only one.
+        """
+        return [
+            {
+                "order_uid": row["order_uid"],
+                "tx_hash": row["tx_hash"],
+                "solver": row["solver"],
+                "data": {
+                    "surplus_fee": str(row["surplus_fee"]),
+                    "amount": float(row["amount"]),
+                    "safe_liquidity": row["safe_liquidity"],
+                },
+            }
+            for row in rewards_df.to_dict(orient="records")
+        ]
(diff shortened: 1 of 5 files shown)


src/models/order_rewards_schema.py
```diff
@@ -0,0 +1,69 @@
...
+                "order_uid": row["order_uid"],
+                "tx_hash": row["tx_hash"],
+                "solver": row["solver"],
+                "data": {
+                    "surplus_fee": str(row["surplus_fee"]),
+                    "amount": float(row["amount"]),
+                    "safe_liquidity": row["safe_liquidity"],
+                },
+            }
+            for row in rewards_df.to_dict(orient="records")
```
bh2smith, line 68: I find all this dataframe manipulation very clunky and I would be open to alternative suggestions here.

The two methods above, for example are not even used in this project because I think two iterations over the same record set is silly and unnecessary. This data class only exists to transform the dataframe into the type we want to stream.

src/sync/order_rewards.py
```diff
@@ -26,37 +26,31 @@ class OrderbookDataHandler(RecordHandler):  # pylint:disable=too-few-public-meth
...
     def write_sync_data(self) -> None:
         # Only write these if upload was successful.
-        config = self.config
-        column = config.sync_column
-        with open(
-            os.path.join(self.file_path, config.sync_file), "w", encoding="utf-8"
-        ) as sync_file:
-            writer = csv.DictWriter(sync_file, fieldnames=[column], lineterminator="\n")
-            writer.writeheader()
-            writer.writerows([{column: str(self.block_range.block_to)}])
```
bh2smith, line 59: One bonus of this change is that this method is now identical to the one for app_data and can be moved into the "super class" (RecordHandler). Unfortunately, the same is not true for `write_found_content`.

src/models/order_rewards_schema.py
bh2smith: Update: I have remove all the unused code: It is here now in case anyone cares

```py

    @classmethod
    def from_pdf(cls, rewards_df: DataFrame) -> list[OrderRewards]:
        """Multi Record Constructor from Pandas Dataframe"""
        return [
            cls(
                order_uid=row["order_uid"],
                tx_hash=row["tx_hash"],
                solver=row["solver"],
                data={
                    "surplus_fee": str(row["surplus_fee"]),
                    "amount": float(row["amount"]),
                    "safe_liquidity": row["safe_liquidity"],
                },
            )
            for row in rewards_df.to_dict(orient="records")
        ]

    def as_dune_record(self) -> dict[str, Any]:
        """Convert to Dune Record"""
        return {
            "order_uid": self.order_uid,
            "tx_hash": self.tx_hash,
            "solver": self.solver,
            "data": self.data,
        }

```

gentrexha: The only update I could think of about this is that you could vectorize it like this:

```python
        rewards_df["data"] = rewards_df.apply(
            lambda row:
                {
                    "surplus_fee": str(row["surplus_fee"]),
                    "amount": float(row["amount"]),
                    "safe_liquidity": row["safe_liquidity"],
                }, axis=1
        )
        return rewards_df[["order_uid", "tx_hash", "solver", "data"]].to_dict(
            orient="records"
        )
```

Let me know what you think about it!

gentrexha (approved):
Good work, Ben!

I left a little comment regarding the df -> records conversion. Other than that, LGTM!

src/models/order_rewards_schema.py
bh2smith: So this is definitely the kind of suggestion I was hoping for (being able to create the "data" column in pandas syntax) however, it looks like this also requires two iterations over the complete data frame (one to build the new column and then one to export the thing in the format). For efficiency reasons, I think I will stick with the current approach. Thanks for this!

PR #21, opened 2022-12-19
Affiliate Data API

bh2smith:
Related to https://github.com/cowprotocol/dune-bridge/issues/52

This PR implements the last change to fully replace the Dune Bridge project.

- A basic API endpoint for user account profiles
- When the address is not a user or an invalid string, default (zero) profile is returned

## TODO

Add more tests (e.g. for API route)

## Test Plan 

Run the API server (requires Dune API key)

Try it out
```
http://127.0.0.1:5000/profile/xyz
127.0.0.1:5000/profile/0x247b317521d7edcfaf9b6d6c21b55217e5c34e0a
```

### DIFF
--- a/src/api/affiliate.py
+++ b/src/api/affiliate.py
@@ -3,6 +3,7 @@ import os
 from typing import Any
 
 from dotenv import load_dotenv
+from dune_client.client import DuneClient
 from flask import Flask, jsonify
 
 from src.fetch.affiliate_data import CachingAffiliateFetcher
@@ -11,7 +12,7 @@ from src.logger import set_log
 app = Flask(__name__)
 load_dotenv()
 cached_fetcher = CachingAffiliateFetcher(
-    api_key=os.environ["DUNE_API_KEY"],
+    dune=DuneClient(os.environ["DUNE_API_KEY"]),
     execution_id=os.environ.get("LAST_EXECUTION_ID"),
     cache_validity=int(os.environ["CACHE_VALIDITY"]),
 )
--- a/src/api/affiliate.py
+++ b/src/api/affiliate.py
@@ -5,12 +5,12 @@ from typing import Any
 from dotenv import load_dotenv
 from flask import Flask, jsonify
 
-from src.fetch.dune_cached import CachedDuneFetcher
+from src.fetch.affiliate_data import CachingAffiliateFetcher
 from src.logger import set_log
 
 app = Flask(__name__)
 load_dotenv()
-cached_fetcher = CachedDuneFetcher(
+cached_fetcher = CachingAffiliateFetcher(
     api_key=os.environ["DUNE_API_KEY"],
     execution_id=os.environ.get("LAST_EXECUTION_ID"),
     cache_validity=int(os.environ["CACHE_VALIDITY"]),
--- a/src/api/affiliate.py
+++ b/src/api/affiliate.py
@@ -1,0 +1,34 @@
+"""Basic Flask API"""
+import os
+from typing import Any
+
+from dotenv import load_dotenv
+from flask import Flask, jsonify
+
+from src.fetch.dune_cached import CachedDuneFetcher
+from src.logger import set_log
+
+app = Flask(__name__)
+load_dotenv()
+cached_fetcher = CachedDuneFetcher(
+    api_key=os.environ["DUNE_API_KEY"],
+    execution_id=os.environ.get("LAST_EXECUTION_ID"),
+    cache_validity=int(os.environ["CACHE_VALIDITY"]),
+)
+log = set_log(__name__)
+
+
+@app.route("/profile/<string:address>", methods=["GET"])
+def get_profile(address: str) -> Any:
+    """
+    Fetches Affiliate Profile for `address`
+    https://api.cow.fi/affiliate/api/v1/profile/<address>
+    """
+    log.info(f"Fetching affiliate data for {address}")
+    affiliate_data = cached_fetcher.get_affiliate_data(account=address)
+    log.debug(f"got results {affiliate_data}")
+    return jsonify(affiliate_data)
+
+
+if __name__ == "__main__":
+    app.run()
--- a/src/api/affiliate.py
+++ b/src/api/affiliate.py
@@ -5,18 +5,18 @@ from typing import Any
 from dotenv import load_dotenv
 from dune_client.client import DuneClient
 from flask import Flask, jsonify
+from flask.logging import create_logger
 
 from src.fetch.affiliate_data import CachingAffiliateFetcher
-from src.logger import set_log
 
 app = Flask(__name__)
+log = create_logger(app)
 load_dotenv()
 cached_fetcher = CachingAffiliateFetcher(
     dune=DuneClient(os.environ["DUNE_API_KEY"]),
     execution_id=os.environ.get("LAST_EXECUTION_ID"),
     cache_validity=int(os.environ["CACHE_VALIDITY"]),
 )
-log = set_log(__name__)
 
 
 @app.route("/profile/<string:address>", methods=["GET"])
@@ -26,8 +26,11 @@ def get_profile(address: str) -> Any:
     https://api.cow.fi/affiliate/api/v1/profile/<address>
     """
     log.info(f"Fetching affiliate data for {address}")
-    affiliate_data = cached_fetcher.get_affiliate_data(account=address)
-    log.debug(f"got results {affiliate_data}")
+    try:
+        affiliate_data = cached_fetcher.get_affiliate_data(account=address)
+    except ValueError as err:
+        return str(err), 400
+
     return jsonify(affiliate_data)
 
 
--- a/src/fetch/affiliate_data.py
+++ b/src/fetch/affiliate_data.py
@@ -58,7 +58,7 @@ class AffiliateMemory:
     last_execution: str
 
 
-class CachedDuneFetcher:
+class CachingAffiliateFetcher:
     """
     Class containing, DuneClient, FileIO and a logger for convenient Dune Fetching.
     """
--- a/src/fetch/affiliate_data.py
+++ b/src/fetch/affiliate_data.py
@@ -88,7 +88,13 @@ class CachingAffiliateFetcher:
             results = self.dune.refresh(Query(AFFILIATE_QUERY_ID), ping_frequency=10)
             log.info(f"query execution {results.execution_id} succeeded")
 
-        last_update = results.times.execution_ended_at or utc_now()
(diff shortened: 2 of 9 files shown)


tests/unit/test_util.py
```diff
@@ -0,0 +1,22 @@
+import unittest
```
gentrexha, line 1: I would have loved to see unit tests testing the `AfflitiateMemory` and `cache_validity` of the `CachingAffiliateFetcher` class.

gentrexha (commented):
LGTM!

Left a comment regarding the unit tests discussion. Other than that, nice work Ben! Although it is worth mentioning, I'm not an expert in Flask.

src/fetch/affiliate_data.py
```diff
@@ -0,0 +1,126 @@
...
+        """Internal method to update the cached affiliate data"""
+        self.memory = self._fetch_affiliates()
+
+    def cache_expired(self) -> bool:
+        """
+        Determines if data cache is expired.
+        """
+        before = self.memory.last_update
+        delta = self.cache_validity
+        return utc_now() > before + delta
```
gentrexha, line 108: Just a discussion for the future since this is a bit late: In my opinion, it's okay for things like these, e.g., cache management, to depend on external packages, i.e., https://github.com/tkem/cachetools

bh2smith: Yea it would be, except that while looking at some of these libraries I noticed that our logic was "too specialized" to use a generic decorator. This way, I feel like we have more specific control. I can look more into it if you think it would serve our purpose out of the box.

tests/unit/test_util.py
```diff
@@ -0,0 +1,22 @@
...
+
+from src.utils import valid_address, utc_now
+
+
+class TestUtils(unittest.TestCase):
```
gentrexha, line 6: In my opinion, we should also be testing that `self._update_memory()` gets called only once if we call `get_affiliate_data` twice within the cache TTL.

gentrexha (approved):
Good job, Ben!

I left a minor test suggestion. Other than that, LGTM!

src/fetch/affiliate_data.py
gentrexha: > I can look more into it if you think it would serve our purpose out of the box.

No need. Just wanted to discuss. Thanks.

tests/unit/test_util.py
bh2smith: Oh yes. We will.

src/fetch/affiliate_data.py
```diff
@@ -0,0 +1,129 @@
...
+        log.info("affiliate data cache loaded")
+
+    def _fetch_affiliates(self, execution_id: Optional[str] = None) -> AffiliateMemory:
+        """Fetches Affiliate data from Dune and updates cache"""
+        if execution_id:
+            results = self.dune.get_result(execution_id)
+        else:
+            # https://dune.com/queries/1757231?d=1
+            # Query executes in ~24 seconds.
+            results = self.dune.refresh(Query(AFFILIATE_QUERY_ID), ping_frequency=10)
```
josojo, line 88: Can this branch end up in an infity_loop? When does refresh stop? Maybe we can add a parameter to refresh that checks how long we are testing? Or is the max time given by the fact that dune will return error if the query runs longer than 30 mins?

src/fetch/affiliate_data.py
```diff
@@ -0,0 +1,129 @@
...
+            log.info(f"query execution {results.execution_id} succeeded")
+
+        last_update = results.times.execution_ended_at or utc_now()
```
josojo, line 91: Why do we  need the or statement? Wouldn't it be better to throw an error?

src/fetch/affiliate_data.py
```diff
@@ -0,0 +1,129 @@
...
+        Returns affiliate data for `account`.
+        If the record on hand is considered "too old",
+        refreshes memory before return.
+        """
+        if not valid_address(account):
+            log.warning(f"Invalid address received {account}")
+
+        if self.cache_expired():
+            log.debug("cache expired - refreshing results")
+            self._update_memory()
```
josojo, line 124: Does this mean that the api query could take 30 secs? We should check with the front-end whether this is okay. 
IF it is not too expensive, we could also just update the cache in the background all 5 minutes and don't update on new requests.

josojo (commented):
Looks good.

One fundamental decision I would like talk about is when we should update the cache. Should we do it in the background continuously or only when the API is pinged.

I think its not nice if a page takes 30 secs to load. Hence, I would prefer to do it in the background

src/fetch/affiliate_data.py
bh2smith: That is a really good point. It technically can take forever except that Dune has a Hard limit at 30 minutes (when the execution status becomes "Failed" and refresh would terminate as well).

src/fetch/affiliate_data.py
bh2smith: I would prefer to update in the background a bit less frequently.

The original Idea I had for this was that we refresh whenever a request comes in AND the cache is expired, but we return the value we currently have. This way we don't need to have scheduled refreshes, but demand triggers an update whenever necessary. Furthermore, nobody would every have to wait because we always return something. I still need to figure out how exactly to implement this, but would prefer to do it in a follow up PR (as it is a more sophisticated feature).

josojo: >The original Idea I had for this was that we refresh whenever a request comes in AND the cache is expired, but we return the value we currently have. This way we don't need to have scheduled refreshes, but demand triggers an update whenever necessary. Furthermore, nobody would every have to wait because we always return something. I still need to figure out how exactly to implement this, but would prefer to do it in a follow up PR (as it is a more sophisticated feature).

But if we would like to ever serve two requests at the same time in parallel, then it could be that both trigger dune to recalculate it, right? Maybe it's not so important, as we will not get there soon. Not sure how we would handle it.
(I don't know how concurrency in python works.)

What do you dislike about my proposal? Is it cost saving?

josojo: I am okay to do it in another PR. sure

bh2smith: Yea... the only thing about the proposal I am not fond of is the 5 minute updates. Would prefer something like 30. I also don't know how to implement the update in parallel with the service at this moment.

src/fetch/affiliate_data.py
bh2smith: This is because `execution_ended_at` is an optional field and we want to have an actual value. Luckily, the field will should always be populated if the query execution was successful. I could throw an error instead if you prefer.

josojo: I think an error would be best practice to know if something did not go well.

bh2smith: Although we would already have errored on the refresh if this field was not available. I will also error here if it happens. Will do.

bh2smith:
I found these docs on concurrency in python 3.11. Will take a look tomorrow. 

https://docs.python.org/3/library/concurrency.html

src/api/affiliate.py
```diff
@@ -0,0 +1,35 @@
...
+from dotenv import load_dotenv
+from dune_client.client import DuneClient
+from flask import Flask, jsonify
+
+from src.fetch.affiliate_data import CachingAffiliateFetcher
+from src.logger import set_log
+
+app = Flask(__name__)
+load_dotenv()
+cached_fetcher = CachingAffiliateFetcher(
```
fleupold, line 14: You should be able to use flask's `APScheduler` to trigger a cache refresh periodically (e.g. once every 5 minutes): https://github.com/viniciuschiele/flask-apscheduler/blob/master/examples/jobs.py

This would allow us to get rid of cache invalidation altogether (which should simplify the code). Basically the cache never becomes invalid it only gets updated periodically (and serves what ever data is current). We should add a prometheus metric on the last updated timestamp and alert if the update wasn't triggered for too long.

This design will also make sure there is only ever one thing writing the cache which makes us have to worry much less about concurrency.

src/fetch/affiliate_data.py
```diff
@@ -0,0 +1,129 @@
...
+        return utc_now() > before + delta
+
+    def get_affiliate_data(self, account: str) -> AffiliateData:
+        """
+        Returns affiliate data for `account`.
+        If the record on hand is considered "too old",
+        refreshes memory before return.
+        """
+        if not valid_address(account):
+            log.warning(f"Invalid address received {account}")
```
fleupold, line 120: Shouldn't we return a 400 error code in this case?

src/api/affiliate.py
```diff
@@ -0,0 +1,35 @@
...
+
+@app.route("/profile/<string:address>", methods=["GET"])
+def get_profile(address: str) -> Any:
+    """
+    Fetches Affiliate Profile for `address`
+    https://api.cow.fi/affiliate/api/v1/profile/<address>
+    """
+    log.info(f"Fetching affiliate data for {address}")
+    affiliate_data = cached_fetcher.get_affiliate_data(account=address)
+    log.debug(f"got results {affiliate_data}")
```
fleupold, line 30: Aren't you double logging this info (once here and once in `get_affiliate_data`)? Let's make sure we log as much as necessary (to debug things) but as little as possible and also follow proper log line format.

Also, aren't we supposed to use app.logger (https://flask.palletsprojects.com/en/1.1.x/logging/)?

src/fetch/affiliate_data.py
bh2smith: Yea we could do that too.

src/api/affiliate.py
bh2smith: I wasn't aware of app logger but will definitely check it out. Hadn't noticed the double logs, but will remove.

src/fetch/affiliate_data.py
bh2smith: I think I would rather set a warning here and use utc now because it makes sense and its not clear what to do otherwise. We definitely don't want the app to crash. I could also log an error (so we know about it), but not crash.

bh2smith: In conclusion I don't think we should "throw" and error, but rather "log" an error (so we are informed, but the app doesn't crash). This would lead us to be notice and report a bug in Dune.

src/fetch/affiliate_data.py
bh2smith: Important to remember that prices may not even become available so frequently.

bh2smith:
Closing because this program has been abandoned. Not sure how we expect to show users their Total Trade Volume in the interface....

PR #23, opened 2023-01-17
[Schema] Add block_number to order_rewards

bh2smith:
For query efficiency reasons, we would need to include block number on the order rewards table in our community sources (this will allow us to join on trades more efficiently). 

⚠️ - this will require coordination with Dune on Redeployment. cc @dsalv

### Test Plan 

Existing Unit tests which have been adapted.

### DIFF
--- a/src/fetch/orderbook.py
+++ b/src/fetch/orderbook.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 import os
 from dataclasses import dataclass
 from enum import Enum
+from typing import Optional
 
 import pandas as pd
 from dotenv import load_dotenv
@@ -46,13 +47,14 @@ class OrderbookFetcher:
 
     @classmethod
     def _read_query_for_env(
-        cls, query: str, env: OrderbookEnv, data_types: dict[str, str]
+        cls, query: str, env: OrderbookEnv, data_types: Optional[dict[str, str]] = None
     ) -> DataFrame:
         return pd.read_sql_query(query, con=cls._pg_engine(env), dtype=data_types)  # type: ignore
 
     @classmethod
-    def _query_both_dbs(cls, query: str) -> tuple[DataFrame, DataFrame]:
-        data_types = {"block_number": "int64", "amount": "float64"}
+    def _query_both_dbs(
+        cls, query: str, data_types: Optional[dict[str, str]] = None
+    ) -> tuple[DataFrame, DataFrame]:
         barn = cls._read_query_for_env(query, OrderbookEnv.BARN, data_types)
         prod = cls._read_query_for_env(query, OrderbookEnv.PROD, data_types)
         return barn, prod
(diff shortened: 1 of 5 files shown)


tests/integration/test_fetch_orderbook.py
```diff
@@ -18,8 +18,10 @@ def test_get_order_rewards(self):
         block_number = 16000000
         block_range = BlockRange(block_number, block_number + 50)
         rewards_df = OrderbookFetcher.get_orderbook_rewards(block_range)
+        print(rewards_df.keys())
```
josojo, line 21: Still wanna have that prints?

src/fetch/orderbook.py
```diff
@@ -45,17 +46,28 @@ def _pg_engine(db_env: OrderbookEnv) -> Engine:
         return create_engine(db_string)
 
     @classmethod
-    def _query_both_dbs(cls, query: str) -> tuple[DataFrame, DataFrame]:
-        barn = pd.read_sql(query, con=cls._pg_engine(OrderbookEnv.BARN))
-        prod = pd.read_sql(query, con=cls._pg_engine(OrderbookEnv.PROD))
+    def _read_query_for_env(
+        cls, query: str, env: OrderbookEnv, data_types: Optional[dict[str, str]] = None
+    ) -> DataFrame:
+        return pd.read_sql_query(query, con=cls._pg_engine(env), dtype=data_types)  # type: ignore
```
bh2smith, line 52: I think there is a bug in mypy on the dtype field. They expect types that I believe match what we pass, but also don't provide a public interface to import their types (also np.int64 doesn't work either).

```
src/fetch/orderbook.py:51: error: Argument "dtype" to "read_sql_query" has incompatible type "Dict[str, str]"; expected "Optional[Union[Union[ExtensionDtype, Union[str, dtype[generic], Type[str], Type[complex], Type[bool], Type[object]]], Dict[Any, Union[ExtensionDtype, Union[str, dtype[generic], Type[str], Type[complex], Type[bool], Type[object]]]]]]"  [arg-type]
```
We ignore types on this line only for now.

josojo: Okay, But I think its better practice to put this comment into the code.

josojo (approved):
Pr look good!

But I don't understand the type issues. Sorry, no background knowledge from my side.

tests/integration/test_fetch_orderbook.py
bh2smith: Will remove

bh2smith:
```suggestion
```

src/fetch/orderbook.py
bh2smith: Done.

bh2smith: I think the comment got too long. I will create an issue to put more context and link it in the comment.

gentrexha (approved):
LGTM!

PR #25, opened 2023-01-19
Add Utility Scripts (empty_bucket & download_file)

bh2smith:
There is no UI to view and manage files in the AWS buckets, so when we want to peek a file we can `download_file`. Furthermore (and more importantly), the script argument `hard-reset` didn't make sense to include in our deployments since we would only every want to do this once. So instead we provided an independent utility script to `empty_bucket` which should be done (one time) before deploying schema changes.

This was already used recently for deployment of the changes to order rewards made in #23.


josojo (approved):
looks good

PR #26, opened 2023-02-15
[Batch Rewards] Orderbook Query for Batch Rewards

bh2smith:
As part of #27 we implement the Raw Orderbook query that will be used to extract the data to be synced with Dune. The follow up to this PR will implement the python script executing this that is also responsible for transforming it into the JSON files being uploaded to Dune's AWS bucket.

Solvers are expected to change their rewards to the following

```
reward(txHash) = observedQuality - referenceScore
```
where `observedQuality = Surplus + Fee`

Furthermore, the reward per batch is planned to be capped by `[-E, E + executionCosts]  (E = 0.01 ETH)`

This query joins several tables in order to provide all terms required to evaluate the reward (namely surplus, fee, execution_cost, reference_score). `winning_score` is included too which is not really necessary but more for transparency.

Furthermore, it is expected that if the total allocated rewards for each accounting period are not reached, then we are to distribute the remaining funds according to the solver participation (found also as a field returned by this query).

For this, we might query something like (but this can come later).

```sql
participations as (
  select 
      tx_hash, 
      unnest(participants) as solver 
  from reward_data
)
select solver,
       count(distinct tx_hash) as num_solutions
from participations
group by solver
```

Trying to add @fhenneke as a reviewer to this, but can't.
 
 
Note that this is all based on the assumption that the following tables exist in the orderbook (introduced in https://github.com/cowprotocol/services/pull/1166)

```sql
CREATE TABLE settlement_scores (
  auction_id bigint PRIMARY KEY,
  winning_score numeric(78,0) NOT NULL,
  reference_score numeric(78,0) NOT NULL
);

CREATE TABLE settlement_observations (
  block_number bigint NOT NULL,
  log_index bigint NOT NULL,
  gas_used numeric(78,0) NOT NULL,
  effective_gas_price numeric(78,0) NOT NULL,
  surplus numeric(78,0) NOT NULL,
  fee numeric(78,0) NOT NULL,

  PRIMARY KEY (block_number, log_index)
);

CREATE TABLE auction_participants (
 auction_id bigint PRIMARY KEY,
 participants bytea[]
);
```

### DIFF
--- a/src/sql/orderbook/batch_rewards.sql
+++ b/src/sql/orderbook/batch_rewards.sql
@@ -1,39 +1,43 @@
-with hashed_observations as (SELECT block_number,
-                                    tx_hash,
-                                    effective_gas_price * gas_used as execution_cost,
-                                    surplus,
-                                    fee,
-                                    auction_id
-                             FROM settlement_observations so
-                                    LEFT OUTER JOIN LATERAL (
-                               SELECT tx_hash, solver, tx_nonce, tx_from
-                               FROM settlements s
-                               WHERE s.block_number = so.block_number
-                                 AND s.log_index = so.log_index
-                               LIMIT 1
-                               ) AS settlement ON TRUE
-                                    JOIN auction_transaction
-                                         ON settlement.tx_from = auction_transaction.tx_from
-                                           AND settlement.tx_nonce = auction_transaction.tx_nonce
-                             WHERE block_number > {{start_block}} AND block_number <= {{end_block}}),
+WITH hashed_observations AS (
+SELECT
+    -- settlement
+    tx_hash,
+    -- settlement_observations
+    block_number,
+    effective_gas_price * gas_used AS execution_cost,
+    surplus,
+    fee,
+    -- auction_transaction
+    auction_id
+FROM settlement_observations so
+JOIN settlements s
+  ON s.block_number = so.block_number
+  AND s.log_index = so.log_index
+JOIN auction_transaction at
+  ON s.tx_from = at.tx_from
+  AND s.tx_nonce = at.tx_nonce
+WHERE block_number > {{start_block}} AND block_number <= {{end_block}}
+),
 
-     reward_data as (select
-                       -- observations
-                       block_number,
-                       tx_hash,
-                       execution_cost,
-                       surplus,
-                       fee,
-                       surplus + fee - reference_score as reward_eth,
-                       -- scores
-                       winning_score,
-                       reference_score,
-                       -- participation
-                       participants
-                     from hashed_observations ho
-                            join settlement_scores ss
-                                 on ho.auction_id = ss.auction_id
-                            join auction_participants ap
-                                 on ho.auction_id = ap.auction_id)
+reward_data AS (
+  SELECT
+     -- observations
+     block_number,
+     tx_hash,
+     execution_cost,
+     surplus,
+     fee,
+     surplus + fee - reference_score AS reward_eth,
+     -- scores
+     winning_score,
+     reference_score,
+     -- participation
+     participants
+FROM hashed_observations ho
+JOIN settlement_scores ss
+  ON ho.auction_id = ss.auction_id
+JOIN auction_participants ap
+  ON ho.auction_id = ap.auction_id
+)
 
-select * from reward_data
+SELECT * FROM reward_data
--- a/src/sql/orderbook/batch_rewards.sql
+++ b/src/sql/orderbook/batch_rewards.sql
@@ -1,0 +1,45 @@
+with
+hashed_observations as (
+  SELECT
+    block_number,
+    tx_hash,
+    effective_gas_price * gas_used as execution_cost,
+    surplus,
+    fee,
+    auction_id
+  FROM settlement_observations so
+  LEFT OUTER JOIN LATERAL (
+    SELECT tx_hash, solver, tx_nonce, tx_from
+    FROM settlements s
+    WHERE s.block_number = so.block_number
+      AND s.log_index > so.log_index
+    ORDER BY s.log_index
+    LIMIT 1
+  ) AS settlement
+    ON TRUE
+  JOIN auction_transaction
+     ON settlement.tx_from = auction_transaction.tx_from
+    AND settlement.tx_nonce = auction_transaction.tx_nonce
+  WHERE block_number > {{start_block}} AND block_number <= {{end_block}}
+)
+
+select
+    -- observations
+    block_number,
+    tx_hash,
+    execution_cost,
+    surplus,
+    fee,
+    surplus + fee - reference_score as reward_eth,
+    -- scores
+    winning_score,
+    reference_score,
+    -- participation
+    participants
+from hashed_observations ho
+-- may want to do outer joins
+join settlement_scores ss
+  on ho.auction_id = ss.auction_id
+-- may want to do outer joins
+join auction_participants ap
+  on ho.auction_id = ap.auction_id
--- a/src/sql/orderbook/batch_rewards.sql
+++ b/src/sql/orderbook/batch_rewards.sql
@@ -2,6 +2,7 @@ WITH observed_settlements AS (
 SELECT
     -- settlement
     tx_hash,
+    solver,
     -- settlement_observations
     block_number,
     effective_gas_price * gas_used AS execution_cost,
@@ -23,6 +24,11 @@ reward_data AS (
   SELECT
     -- observations
     tx_hash,
+    coalesce(
+      solver,
+      -- This is the winning solver (i.e. last entry of participants array)
+      participants[array_length(participants, 1)]
+    ) as solver,
     -- Right-hand terms in coalesces below represent the case when settlement
     -- observations are unavailable (i.e. no settlement corresponds to reported scores).
     -- In particular, this means that surplus, fee and execution cost are all zero.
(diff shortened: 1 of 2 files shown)


src/sql/orderbook/batch_rewards.sql
```diff
@@ -0,0 +1,40 @@
...
+                       -- scores
+                       winning_score,
+                       reference_score,
+                       -- participation
+                       participants
+                     from hashed_observations ho
+                            join settlement_scores ss
+                                 on ho.auction_id = ss.auction_id
+                            join auction_participants ap
+                                 on ho.auction_id = ap.auction_id)
```
bh2smith, line 38: We may want to perform outer joins here so we can tell if something is missing. However, it would be much better if the order-book could provide some guarantees that things will be there (since the db schema itself cannot).

src/sql/orderbook/batch_rewards.sql
```diff
@@ -0,0 +1,40 @@
...
+                                    fee,
+                                    auction_id
+                             FROM settlement_observations so
+                                    LEFT OUTER JOIN LATERAL (
+                               SELECT tx_hash, solver, tx_nonce, tx_from
+                               FROM settlements s
+                               WHERE s.block_number = so.block_number
+                                 AND s.log_index > so.log_index
+                               ORDER BY s.log_index
+                               LIMIT 1
```
gentrexha, line 14: I'm just wondering: Are you trying to find all settlement_observations that have a matching row in the settlement table?

src/sql/orderbook/batch_rewards.sql
gentrexha: Good point. I think we shouldn't take these things for granted and still test for them.

gentrexha (approved):
Left one comment for discussion.

Other than that, LGTM!

src/sql/orderbook/batch_rewards.sql
```diff
@@ -0,0 +1,40 @@
+with hashed_observations as (SELECT block_number,
+                                    tx_hash,
+                                    effective_gas_price * gas_used as execution_cost,
+                                    surplus,
+                                    fee,
+                                    auction_id
+                             FROM settlement_observations so
+                                    LEFT OUTER JOIN LATERAL (
```
gentrexha, line 8: Also, #TIL about `LATERAL`

src/sql/orderbook/batch_rewards.sql
```diff
@@ -0,0 +1,40 @@
...
+                               SELECT tx_hash, solver, tx_nonce, tx_from
+                               FROM settlements s
+                               WHERE s.block_number = so.block_number
+                                 AND s.log_index > so.log_index
```
sunce86, line 12: Why is `>` here?

sunce86 (commented):
Lg, I am just wondering for the `log_index`.

src/sql/orderbook/batch_rewards.sql
bh2smith: Settlement event is logged after all trades have gone through at the end of the transaction (as the last event). It is technically possible that two batches are settled in the same transaction so we need to match the correct trades with the correct settlements.

sunce86: Oh, you expect `log_index` in `settlement_observations` to be index of the trade event, not the index of settlement index?

src/sql/orderbook/batch_rewards.sql
gentrexha: I'm unsure how much performance dependent we are on this query, but using a `WHERE EXISTS` might be better for performance than a join.

src/sql/orderbook/batch_rewards.sql
bh2smith: No.. I expect it to be the log index of the settlement event of course. I mean this is a pretty common join being made... but maybe I have misinterpreted.

Maybe we should cross reference with this.

https://github.com/cowprotocol/services/blob/ac5ac7365a03ffb44a44d0c840b16396f66c9fb6/crates/database/src/trades.rs#L28-L48

In the case of this query -- we aren't actually joining on trade events so perhaps the fancy join isn't necessary. Would be a lot easier to validate this if there were already populated data...

bh2smith: Actually this may not be necessary. What matters to me here is that we need a SINGLE `tx_hash` for a batch but it is possible that there are two settlements in the same transaction. I really don't know how to deal with this just yet, but I do think that maybe this query needs to be tested and fixed.

bh2smith: Maybe we don't need to do the lateral join at all. Problem is this will all break if there are two settlement in the same transaction.

bh2smith: OK I see the issue here. We should not be using inequality, but rather equality on the log index. Sorry for the confusion (this query was partially copied from the join on trades above).

bh2smith: It is very good that you have pointed this out. I would like to get some more eyes on this. How will the services handle the scenario when there are two settlements in the same tx?

sunce86: Cross reference you mentioned is indeed related to trade events that have their own `log_index` which is different (lower) than settlement `log_event` so I _think_  this does not apply in our case, where we simply use the same event in both tables.

> How will the services handle the scenario when there are two settlements in the same tx?

Have no idea, will think about it.

bh2smith: Ok, so I have dramatically simplified this query for the time being and would propose we assume this is good for now and circle back to fix it as necessary once the data actually becomes available.

sunce86 (approved):

src/sql/orderbook/batch_rewards.sql
```diff
@@ -0,0 +1,44 @@
...
+    -- settlement
+    tx_hash,
+    -- settlement_observations
+    block_number,
+    effective_gas_price * gas_used AS execution_cost,
+    surplus,
+    fee,
+    -- auction_transaction
+    auction_id
+FROM settlement_observations so
```
harisang, line 12: I guess this is a question for the PR that introduced the `settlement_observations` table, but just wanted to clarify here. Does this also contain the settlements that reverted?

bh2smith: Not to my knowledge, how could it? There would be nothing to be observed, but this is a great point. we can get that information from dune, but I think there is a tendency to try and be able to compute this entirely from our own db. In that case one would have to monitor all call to the contract (I.e failed transactions) and this would have to be part of the backend services which it DEFINITELY is not currently.

harisang: >  but I think there is a tendency to try and be able to compute this entirely from our own db. In that case one would have to monitor all call to the contract (I.e failed transactions) and this would have to be part of the backend services which it DEFINITELY is not currently.

Yes, I would expect that we do everything from our own db. And to me, although I might be completely wrong here, it makes sense that for every auction id that has a winning solution, we store all onchain details around it, if any. This would imply storing a tx hash (and whatever relevant information we need) even in the case of a revert, as this connects our offchain auction with the corresponding onchain activity, if any.

src/sql/orderbook/batch_rewards.sql
```diff
@@ -0,0 +1,44 @@
...
+
+reward_data AS (
+  SELECT
+     -- observations
+     block_number,
+     tx_hash,
+     execution_cost,
+     surplus,
+     fee,
+     surplus + fee - reference_score AS reward_eth,
```
fhenneke, line 30: to be consistent with the proposal, maybe rename `reward_eth` to something like `payment`

src/sql/orderbook/batch_rewards.sql
sunce86: For reverted transactions we would populate only `block_number, effective_gas_price, gas used` right?
Is this MUST HAVE for CIP20? Backend relies on emitted events so currently there is no way to index reverted transactions, implementing this would probably delay CIP20 for a few days. I would rather do it as a follow up.

bh2smith: As discussed earlier today, it would suffice to add a column `block_deadline` to the settlement_scores table for which we can infer these non-existent values and still be able to assign a block number to the batch.

src/sql/orderbook/batch_rewards.sql
bh2smith:
```suggestion
     surplus + fee - reference_score AS payment,
```

bh2smith:
Ok so this query now contains all the requirements (assuming several things about the backend).

1. all valuations are in ETH (wei)
2. winning solver is the last element of the participants array in the auction_participants table
3. there will always be a unique record linking settlement scores and auction_participants, but not always a corresponding settlement
4. block_deadline exists.

Note that this query may not actually be fully syntactically correct, but can be fixed once this data exists.
cc @harisang and/or @fhenneke for one last look through, but will be merging sometime today.

Here is a test query on a postgres DB extracting the last element from a byte array

```sql
drop table if exists test;
create table test
(
  my_array bytea[]
);
insert into test (my_array)
values (ARRAY ['\x1212'::bytea, '\x2312'::bytea, '\x3412'::bytea]);
select
    my_array,
    array_length(my_array, 1) as arr_length
    my_array[array_length(my_array, 1)] as last_element
from test;
```

Issue #27, opened 2023-02-15
[Epic] Solver Auction CIP


bh2smith:
Solvers are changing their reward structure (yet again) and we will need to adapt what information is synced.

- [x] Create necessary tables in Orderbook Database (https://github.com/cowprotocol/services/pull/1166)
- [ ] Query for info to be streamed into Dune Community Sources (https://github.com/cowprotocol/dune-sync/pull/26)
- [ ] Construct appropriate Dune Schema for Batch Rewards.

bh2smith:
Since I don't think its very appropriate to create even a draft PR in an external repo before we are prepared to utilize it, I will prepare the "spell" code here for later use.


```sql
{{ config(alias='batch_rewards',
        post_hook='{{ expose_spells_hide_trino(\'["ethereum"]\',
                                    "project",
                                    "cow_protocol",
                                    \'["bh2smith"]\') }}'
)}}

-- PoC Query here - <PROVIDE QUERY URL>
select
    distinct tx_hash,
    block_number,
    solver,
    data.reward_eth as unbound_reward,
    data.execution_cost,
    data.participants
from {{ source('cowswap', 'raw_batch_rewards') }}
```



From this we would write a parameterized query to dynamically impose reward capping as follows  

```sql
select 
    tx_hash, 
    solver, 
    min(max(-{{EPSILON}}, unbound_reward), {{EPSILON}} + execution_cost) 
from batch_rewards
```

For participation distribution we would build another query that "unnests" or "explodes" the participants array and transposes.

```sql
with expanded_participation as (
    select 
        block_number,
        tx_hash,
        explode(participants) as participant
    from batch_rewards
)

select 
    participant, 
    count(*) as num_solutions 
from expanded_participation
```

PR #28, opened 2023-02-15
[Batch Rewards] Schema & ETL Process for Sync

bh2smith:
Another part of #27 (based on #26)

This defined the "schema" for what will be uploaded to Dune. We reused/generalize some common code for the order rewards sync. More or less, this was a minor change. Still remaining todo for parsing the `bytea[]` type with participants field (since duneV2 uses lower case hex strings). We will want to replace all `\x` coming from postgres with `0x`


src/fetch/orderbook.py
```diff
@@ -90,3 +90,20 @@ def get_orderbook_rewards(cls, block_range: BlockRange) -> DataFrame:
...
+        cow_reward_query = (
+            open_query("orderbook/batch_rewards.sql")
+            .replace("{{start_block}}", str(block_range.block_from))
+            .replace("{{end_block}}", str(block_range.block_to))
+        )
+        data_types = {"block_number": "int64"}
+        barn, prod = cls._query_both_dbs(cow_reward_query, data_types)
+
+        # Solvers do not appear in both environments!
+        assert set(prod.solver).isdisjoint(set(barn.solver)), "solver overlap!"
```
gentrexha, line 108: This seems a lot more straightforward than the comparison we were doing before with data frames. Good job!

src/sync/orderbook.py
```diff
@@ -0,0 +1,125 @@
...
+        self.data_list = data_list
+
+    def num_records(self) -> int:
+        return len(self.data_list)
+
+    def write_found_content(self) -> None:
+        self.file_manager.write_ndjson(data=self.data_list, name=self.content_filename)
+
+    def write_sync_data(self) -> None:
+        # Only write these if upload was successful.
```
gentrexha, line 49: Is there a check for this?

bh2smith: Not sure I understand the question. Are you asking if this method is tested somewhere?

bh2smith: It might be worth mentioning that this file was just renamed from `order_rewards.py` because the content is almost completely reused. Maybe its better if I don't perform the rename here so we do not need to look at logic that has not changed. I do believe this method has been tested, but we may have to double check.

src/sync/orderbook.py
```diff
@@ -0,0 +1,125 @@
...
+    sync_orderbook_data(
+        aws,
+        block_range,
+        config,
+        dry_run,
+        sync_table,
+        data_list=BatchRewards.from_pdf_to_dune_records(
+            fetcher.get_order_rewards(block_range)
+        ),
+    )
```
gentrexha, line 103: In my opinion, there's a lot of duplicated code across these two methods. Maybe this can benefit from one more parameter to be either batch or orders? Although, readability and understandability will probably take a hit doing that.

What do you think?

gentrexha (approved):
Left one discussion comment and one minor question.

Other than that, LGTM!

src/sync/orderbook.py
gentrexha: From the comment, I understood that this code below is executed only if the upload is successful. I wanted to know if you do check somewhere if the upload was successful because I couldn't find it.

src/sync/orderbook.py
bh2smith: I have already tried quite excessively to split up the duplicated components of this code, but I could try again one more time. At this point, the only thing I can see that is left to deduplicate is the fetching of `block_range` and then maybe creating a general interface for these two methods 

- `BatchRewards.from_pdf_to_dune_records`
- `OrderRewards.from_pdf_to_dune_records`

In fact, looking here at this code you even helped me to spot a bug which I will now highlight as a suggested code change.

```suggestion
def sync_order_rewards(
    aws: AWSClient, fetcher: OrderbookFetcher, config: SyncConfig, dry_run: bool
) -> None:
    """Order Rewards Data Sync Logic"""
    sync_table = SyncTable.ORDER_REWARDS
    block_range = BlockRange(
        block_from=last_sync_block(
            aws,
            table=sync_table,
            genesis_block=15719994,  # First Recorded Order Reward block
        ),
        block_to=fetcher.get_latest_block(),
    )
    sync_orderbook_data(
        aws,
        block_range,
        config,
        dry_run,
        sync_table=sync_table,
        data_list=OrderRewards.from_pdf_to_dune_records(
            fetcher.get_order_rewards(block_range)
        ),
    )


def sync_batch_rewards(
    aws: AWSClient, fetcher: OrderbookFetcher, config: SyncConfig, dry_run: bool
) -> None:
    """Batch Reward Sync Logic"""
    sync_table = SyncTable.BATCH_REWARDS
    block_range = BlockRange(
        block_from=last_sync_block(
            aws,
            table=sync_table,
            genesis_block=15719994,  # First Recorded Order Reward block
        ),
        block_to=fetcher.get_latest_block(),
    )
    sync_orderbook_data(
        aws,
        block_range,
        config,
        dry_run,
        sync_table,
        data_list=BatchRewards.from_pdf_to_dune_records(
            fetcher.get_batch_rewards(block_range)
        ),
    )
```

gentrexha: I think that's a great idea. And glad you found that, although I'm not quite sure what was replaced with what.

bh2smith: RIght. I guess you can't see it because of the file rename. I will revert the file rename so you can see what really happened here. Sorry about that.

bh2smith: Ok --- you will find all this stuff now under the original  file name `order_rewards.py` with a proper DIFF in the code.

src/sync/order_rewards.py
```diff
@@ -56,23 +56,70 @@ def write_sync_data(self) -> None:
...
+    aws: AWSClient, fetcher: OrderbookFetcher, config: SyncConfig, dry_run: bool
+) -> None:
+    """Batch Reward Sync Logic"""
+    sync_table = SyncTable.BATCH_REWARDS
     block_range = BlockRange(
         block_from=last_sync_block(
             aws,
-            table=SYNC_TABLE,
+            table=sync_table,
             genesis_block=15719994,  # First Recorded Order Reward block
```
fleupold, line 90: Shouldn't this be a different block?

src/fetch/orderbook.py
```diff
@@ -75,7 +75,7 @@ def get_latest_block(cls) -> int:
         return min(int(barn["latest"][0]), int(prod["latest"][0])) - REORG_THRESHOLD
 
     @classmethod
-    def get_orderbook_rewards(cls, block_range: BlockRange) -> DataFrame:
+    def get_order_rewards(cls, block_range: BlockRange) -> DataFrame:
```
fleupold, line 78: method comment should probably be updated to reflect that this is the old reward scheme.

fleupold (commented):
How is this supposed to be run? As another pod running the same python script but with different arguments? Or is the same pod going to sync both order as well as batch rewards (we will need both for the transition period).

Some more comments/questions in line...

src/models/batch_rewards_schema.py
```diff
@@ -0,0 +1,38 @@
...
+    @classmethod
+    def from_pdf_to_dune_records(cls, rewards_df: DataFrame) -> list[dict[str, Any]]:
+        """Converts Pandas DataFrame into the expected stream type for Dune"""
+        return [
+            {
+                "block_number": int(row["block_number"]),
+                "tx_hash": row["tx_hash"],
+                "solver": row["solver"],
+                "data": {
+                    "reward_eth": float(row["reward_eth"]),
```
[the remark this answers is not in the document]
bh2smith: Currently the per order rewards are in COW as floats because that is how we receive them from the DB. I suppose this was just a copy-pasta issue since the setup is so similar.

src/models/batch_rewards_schema.py
```diff
@@ -0,0 +1,38 @@
...
+                    # All the following values are in WEI.
+                    "execution_cost": str(row["execution_cost"]),
+                    "surplus": str(row["surplus"]),
+                    "fee": str(row["fee"]),
+                    "winning_score": str(row["winning_score"]),
+                    "reference_score": str(row["reference_score"]),
+                    # TODO - Not sure yet how to parse this bytea[]
+                    #  Will need to experiment with this.
+                    "participants": row["participants"],
```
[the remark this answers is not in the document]
bh2smith: I think its fine for us to just use whatever we want here and it can just remain the same in the back-end.

src/sql/orderbook/batch_rewards.sql
```diff
@@ -21,11 +22,12 @@ with hashed_observations as (SELECT block_number,
...
                        -- observations
                        block_number,
-                       tx_hash,
+                       concat('0x', encode(tx_hash, 'hex'))  as tx_hash,
+                       concat('0x', encode(solver, 'hex'))  as solver,
                        execution_cost,
                        surplus,
                        fee,
-                       surplus + fee - reference_score as reward_eth,
+                       (surplus + fee - reference_score) / pow(10, 18) as reward_eth,
```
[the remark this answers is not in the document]
bh2smith: We can keep this in WEI -- its just that we have had issues with precision and we were already doing this for COW. I can fix this since we do actually still have WEI. In that case we may have to cast this to `:: text`

src/sync/order_rewards.py
bh2smith: Yes absolutely! Great catch (copy-pasta).

bh2smith:
> How is this supposed to be run? As another pod running the same python script but with different arguments? Or is the same pod going to sync both order as well as batch rewards (we will need both for the transition period).

Initially this has been setup to run as another pod with different runtime arguments, but now that you mention it, we don't really need to do that since it will be phased out so soon. Perhaps we can just put both orderbook table syncs into the same job (they execute almost instantly and, use almost no computational resources and do nearly the same thing). Then later we can just drop the other job from here and version bump the production deployment!

src/sql/orderbook/batch_rewards.sql
```diff
@@ -0,0 +1,42 @@
...
+
+     reward_data as (select
+                       -- observations
+                       block_number,
+                       concat('0x', encode(tx_hash, 'hex'))  as tx_hash,
+                       concat('0x', encode(solver, 'hex'))  as solver,
+                       execution_cost,
+                       surplus,
+                       fee,
+                       (surplus + fee - reference_score) / pow(10, 18) as reward_eth,
```
harisang, line 30: How do we convert `surplus` and `fees` into ETH here?

bh2smith: I was told that everything was already in ETH and that is a fundamental assumption we make of the orderbook. Cc @fhenneke

src/sync/orderbook.py
bh2smith: Upload being successful is reaching the next point in the linearly executed code. If the upload was not successfull the script run would fail.

src/sync/order_rewards.py
bh2smith: We may not actually know this block till later but its really just any lower bound so I will choose a recent block from today for now and leave a TODO here.

bh2smith:
This PR is getting pretty old. All comments have been addressed and I plan to merge this today.

src/models/batch_rewards_schema.py
```diff
@@ -0,0 +1,38 @@
...
+    def from_pdf_to_dune_records(cls, rewards_df: DataFrame) -> list[dict[str, Any]]:
+        """Converts Pandas DataFrame into the expected stream type for Dune"""
+        return [
+            {
+                "block_number": int(row["block_number"]),
+                "tx_hash": row["tx_hash"],
+                "solver": row["solver"],
+                "data": {
+                    "reward_eth": str(row["reward_eth"]),
+                    # All the following values are in WEI.
```
fleupold, line 26: nit: move up one line

fleupold (approved):

bh2smith:
Merging now as the general sentiment is correct. There have been some changes to the backend DB that will need to be addressed as a follow up once their structures have been deployed and populated.

PR #33, opened 2023-03-24
[Bug] Fix numerical cast overflow

bh2smith:
Sync has been failing the last 24 hours with the following error:

```
  File "pandas/_libs/lib.pyx", line 2411, in pandas._libs.lib.maybe_convert_numeric
ValueError: Integer out of range. at position 13
```

Turns out this numerical cast was never actually necessary and only resulted in subtle precision loss. 

This does not affect the format of the JSON content being uploaded to the AWS buckets.

## Test Plan

With (DB credentials set) run dry run mode

```
python -m src.main --sync-table  batch_rewards --dry-run True
```

and see logs like

```
INFO src.sync.order_rewards Handling 1438 new records
INFO src.sync.upload_handler attempting to post 1438 new batch_rewards records for block range BlockRange(from=16891896, to=16903266)
INFO src.sync.upload_handler DRY-RUN-ENABLED: New records written to volume, but not posted to AWS.
INFO src.sync.order_rewards batch_rewards sync run completed successfully
```

instead of the Value error displayed above.


fhenneke:
As far as I can tell the changes look good.
Question: Does a change of our internal representation for the data imply that the data on Dune will change as well?

fhenneke (approved):

bh2smith:
> Question: Does a change of our internal representation for the data imply that the data on Dune will change as well?

What we upload to dune should remain be identical.

PR #37, opened 2023-04-27
[Slippage V2] Sync Internal Transfers

bh2smith:
Adding `internal_imbalance` sync script, which is basically the same as the orderbook jobs, but reading from a different DB. The query is not overly complicated making a single join to get block numbers.

Note that we wrote a `PostgresFetcher` class to read from this DB (since the OrderbookFetcher reads from two DBs and combines results). More natural approach would be to have a `MultiInstanceDBFetcher` recently introduced in solver-rewards [here](https://github.com/cowprotocol/solver-rewards/pull/200/files#diff-14bc0f308aa5dcfb330f143b5cfe3fe34dd01df48d289adb8b6277d065df3994)

## Test Plan

new tests introduced

- [x] Have already done a full e2e test writing to their staging buckets.

### DIFF
--- a/tests/e2e/test_sync_app_data.py
+++ b/tests/e2e/test_sync_app_data.py
@@ -46,6 +46,7 @@ class TestSyncAppData(IsolatedAsyncioTestCase):
             block_range=block_range_1,
             config=self.config,
             missing_file_name=missing_files,
+            ipfs_access_key=os.environ["IPFS_ACCESS_KEY"],
         )
 
         print(f"Beginning Content Fetching on {len(data_handler.new_rows)} records")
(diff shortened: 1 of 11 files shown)


tests/e2e/test_sync_app_data.py
```diff
@@ -46,6 +46,7 @@ async def test_fetch_content_and_filter(self):
             block_range=block_range_1,
             config=self.config,
             missing_file_name=missing_files,
+            ipfs_access_key=os.environ["IPFS_ACCESS_KEY"],
```
bh2smith, line 49: This was missed in a previous PR... e2e tests aren't run in our CI so it was not noticed.

PR #38, opened 2023-05-01
Add Log Configuration (directed to stdout)

### LINKED ISSUE #36: All logs are going to `stderr`
`dune-sync` pods are currently flooding `#alerts-prod` because all logs are going to `stderr`.
This can be seen in our logs [here](https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-7d%2Fd,to:now))&_a=(columns:!(log),filters:!(),grid:(columns:('@timestamp':(width:164))),index:c0e240e0-d9b3-11ed-b0e6-e361adffce0b,interval:auto,query:(language:kuery,query:'kubernetes.pod_name:%20%22order-rewards-cronjob%22%20and%20stream:%20stderr'),sort:!(!('@timestamp',desc)))).
This is true for `batch-rewards`, `app-data`, `app-data` and probably all `dune-sync` pods.

bh2smith:
Closes #36 

Will highlight the key part of this change inline.


logging.conf
```diff
@@ -0,0 +1,28 @@
...
+level=DEBUG
+qualname=src
+handlers=consoleHandler
+propagate=0
+
+[handler_consoleHandler]
+class=StreamHandler
+level=DEBUG
+formatter=sampleFormatter
+args=(sys.stdout,)
```
bh2smith, line 25: StreamHandler being configured as `sys.stdout`).

PR #44, opened 2023-06-09
[AppData] Support for New Format

bh2smith:
This change extends the Cid class to support arbitrary prefixes and explicitly makes use of both the OLD and NEW format that we use in https://github.com/cowprotocol/services. 

Because we will need to support both for eternity, the app content fetching is reworked in a way that prioritizes the new format and only attempts the old format when the new one is unavailable.

I believe that we could completely remove support for the old format if all the existing files were migrated to the corresponding CIDs (i.e. those constructed with the new format) -- cc @vkgnosis is this something we plan to do? If we did this, then the changes to be made here would be only a one line change of prefix.

Note this is based on (and blocked by #45)

### DIFF
--- a/src/fetch/ipfs.py
+++ b/src/fetch/ipfs.py
@@ -6,6 +6,7 @@ from typing import Any, Optional
 
 import aiohttp
 import requests
+from aiohttp import ClientSession
 from multiformats_cid.cid import from_bytes
 
 from src.logger import set_log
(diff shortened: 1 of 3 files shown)


vkgnosis:
> I believe that we could completely remove support for the old format if all the existing files were migrated to the corresponding CIDs (i.e. those constructed with the new format)

It's not possible. If an order's app data was set to the hash obtained from the old method, then we cannot migrate that to refer to the same (or any) document when interpreted as the new method.

src/fetch/ipfs.py
```diff
@@ -85,6 +93,12 @@ async def fetch_many(  # pylint: disable=too-many-locals
                 result = await cid.fetch_content(
                     max_retries, previous_attempts, session, first_seen_block
                 )
+                if isinstance(result, NotFoundContent):
+                    # Try Fetching the old format
+                    result = await cls.old_schema(app_hash).fetch_content(
+                        max_retries, previous_attempts, session, first_seen_block
```
vkgnosis, line 99: You could do it in parallel instead of only when the first try fails.

vkgnosis (approved):

gentrexha (approved):
Left one nitpick comment. Other than that LGTM.

tests/unit/test_ipfs.py
```diff
@@ -15,20 +15,30 @@ def test_cid_parsing(self):
...
 
+    def test_new_hash_to_cid(self):
+        self.assertEqual(
+            "bafkrwiek6tumtfzvo6yivqq5c7jtdkw6q3ar5pgfcjdujvrbzkbwl3eueq",
+            str(
+                Cid(
+                    "0x8af4e8c9973577b08ac21d17d331aade86c11ebcc5124744d621ca8365ec9424"
+                )
+            ),
+        )
```
bh2smith, line 40: @gentrexha - this is a test of the new schema!

And it is the same test as the services project uses: see [here](https://github.com/cowprotocol/services/blob/2db800aa38824e32fb542c9e2387d77ca6349676/crates/app-data-hash/src/lib.rs#L63-L76)

tests/unit/test_ipfs.py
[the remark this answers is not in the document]
bh2smith: We did add a test: 

https://github.com/cowprotocol/dune-sync/pull/44/files#r1226432296

tests/unit/test_ipfs.py
gentrexha: Thanks a lot for pointing this out! I must've missed it.

gentrexha (approved):

bh2smith:
> Thanks a lot for pointing this out! I must've missed it.

It was easy to miss in a sea of other tests being changed to test the old style.

tests/unit/test_ipfs.py
bh2smith: It was easy to miss in a sea of other tests being changed to test the old style.

src/fetch/ipfs.py
bh2smith: I made an issue for this:

https://github.com/cowprotocol/dune-sync/issues/49

bh2smith:
Drafted a release to go with this change. We can publish the release just a bit before this new app data goes into staging.

PR #53, opened 2023-06-21
[AWS] Get Credentials from Session

bh2smith:
In an effort to supply fewer env vars to our pods, when the image is deployed on AWS, certain session tokens for obtaining credentials are already injected into the runtime environment. 

FYI (@cowprotocol/analytics)
This is an experimental PR made in collaboration with @mateipopa from devops that will be reverted if it does not work out. Alternatively we could deploy a beta version based on this branch and test that out in staging before even merging.

### DIFF
--- a/src/post/aws.py
+++ b/src/post/aws.py
@@ -138,16 +138,13 @@ class AWSClient:
                 RoleSessionName="InternalSession",
             )
             credentials = internal_assumed_role_object["Credentials"]
-        else:
-            # When deployed to AWS and
-            credentials = sts_client.get_session_token()["Credentials"]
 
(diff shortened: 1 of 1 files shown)


gentrexha (approved):

mateipopa (approved):

bh2smith:
@mateipopa -- here is a pre-release deployment:

https://github.com/cowprotocol/dune-sync/releases/tag/v0.5.5-beta

bh2smith:
So we tried this out and it works like a dream now. WIll merge to main and then this other change can be reverted: https://github.com/cowprotocol/infrastructure/pull/477#discussion_r1238873201

PR #54, opened 2023-07-04
[WIP] CIP-27: add quote solver to order rewards

bh2smith:


fhenneke:
One thing I would add is an e2e test for a more recent block range. At the moment we only test a block range with missing quoting solvers.

Otherwise LGTM.

fhenneke (approved):

PR #55, opened 2023-07-06
Fetch appdata from backend

MartinquaXD:
It's possible that in the future some app data content will only be available in the cow protocol backend.
To make the app data indexing work properly in those cases we need to add another fallback to app data fetching.
The new order is:
1. try new format on IPFS
2. try old format on IPFS
3. try any format on the prod backend
4. try any format on the staging backend

### Tests
* unit test that verifies that content that `fetch_from_backend()` is able to find content that only exists on prod or only exists on prod
* unit test that verifies that hex strings with or without `0x` prefix
* a unit test that verifies that `fetch_many()` is able to fetch app data that only exists in the backend was not possible at the time because such data didn't exist yet

### DIFF
--- a/src/fetch/ipfs.py
+++ b/src/fetch/ipfs.py
@@ -39,21 +39,14 @@ class Cid:
     ) -> FoundContent | NotFoundContent:
         """Fetches the given app data hash from the cow protocol backend (prod and staging)"""
 
-        # first try prod
-        url = f"https://api.cow.fi/mainnet/api/v1/app_data/{hex_str}"
-        response = cls.fetch_from_backend_inner(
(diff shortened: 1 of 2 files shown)


MartinquaXD:
cc @vkgnosis

vkgnosis (approved):

nlordell (approved):

gentrexha (approved):
LGTM. Left a nitpicky comment.

bh2smith (approved):
Good stuff. should we be tagging this asap?

MartinquaXD:
> should we be tagging this asap?

I think deploying this before we release the custom interactions feature next Tuesday is enough.

PR #56, opened 2023-07-13
[Script] Delete From {{BlockNumber}}

bh2smith:
Recently we wanted to empty only part of the bucket (namely all files newer than a certain block number). 
This script does the thing.

### DIFF
--- a/src/post/aws.py
+++ b/src/post/aws.py
@@ -251,3 +251,20 @@ class AWSClient:
             raise ValueError(
                 f"Invalid table_name {table}, please chose from {SyncTable.supported_tables()}"
             ) from err
+
+    def delete_from(self, table: SyncTable | str, block_number: int) -> None:
+        """Deletes all files above and including `block_numer`"""
+        log.info(f"Deleting all files in {table} from {block_number}")
+        try:
+            table_files = self.existing_files().get(table)
+            filtered_files = [
+                fd for fd in table_files if fd.block is None or fd.block >= block_number
+            ]
+            log.info(f"Found {len(filtered_files)} files to be removed.")
+            for file_data in filtered_files:
+                log.info(f"Deleting file {file_data.object_key}")
+                self.delete_file(file_data.object_key)
+        except KeyError as err:
+            raise ValueError(
+                f"Invalid table_name {table}, please chose from {SyncTable.supported_tables()}"
+            ) from err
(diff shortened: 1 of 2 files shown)


src/fetch/ipfs.py
```diff
@@ -129,18 +129,18 @@ async def fetch_many(  # pylint: disable=too-many-locals
                     max_retries, previous_attempts, session, first_seen_block
                 )
```
MartinquaXD, line 130: I thought the plan was to first look in the backend?

MartinquaXD (approved):
Not super familiar with the code or even python but the PR looks reasonable. 👌

src/fetch/ipfs.py
bh2smith: I just switched around trying to fetch backend with trying to fetch old schema, then realized it didn't even help.

src/post/aws.py
```diff
@@ -251,3 +251,20 @@ def delete_all(self, table: SyncTable | str) -> None:
             raise ValueError(
                 f"Invalid table_name {table}, please chose from {SyncTable.supported_tables()}"
             ) from err
+
+    def delete_from(self, table: SyncTable | str, block_number: int) -> None:
+        """Deletes all files above `block_numer`"""
```
gentrexha, line 256: Nit:
```suggestion
        """Deletes all files above and including `block_numer`"""
```

tests/unit/test_ipfs.py
```diff
@@ -87,6 +87,14 @@ def test_get_content(self):
...
         )
 
+    def test_get_new_content(self):
+        x = "0x22e58ea20cbe7269450637ffc0f40fee7bca0d66fc6bf63666bae4ba08c99688"
+        y = "0x93fc55a145fafb4e65a55f92ea416b55be4b4720c2a2c5c0071b67800c6235b2"
+        z = "0xe0e1f18b98680421df973d2d92d0f1603d35dcad58890f2f9de6ac914bcd54ae"
+        for element in [x, y, z]:
+            content = Cid.fetch_from_backend(element, 1, 10)
+            print(content)
+
```
gentrexha, line 97: Can you please elaborate how this is relevant to the PR?

gentrexha (approved):
Left some discussion comments. LGTM.

tests/unit/test_ipfs.py
bh2smith: It isn't, I was testing something else in the same place (when I realized that I needed this new script). Will clean up before merge. The only thing we need here is the delete_from script.

PR #57, opened 2023-07-14
[AppData] Try Fetch from backend first

bh2smith:
This makes things a lot faster. Ideally we can even write a DB query to determine which "new" app hashes we should be fetching (by first_seen block). This would cost us SO much less in Dune fees (because now we query dune for new appHashes).


fleupold:
Superseded by #103

Issue #62, opened 2024-01-10
The syncing of quoting data is now failing due to changes in our database.


harisang:
Otex mentioned this morning that we are not syncing which solver quoted each order anymore on Dune, and this seems to have started after yesterday's weekly release was applied.

Not sure if this is only due to the following, but at least one change is in the `order_execution` table, where we now only have entries for limit orders but not market orders. This means we need to adjust the queries here in order to properly recover the quoter of each order and continue pushing this information on Dune.

PR #63, opened 2024-01-10
Fix query that computes quoter and syncs data on Dune

harisang:
This PR addresses issue #62, which was caused by a change in our database. Specifically, the `order_execution` table now only contains entries for limit orders (and NOT market orders).

### DIFF
--- a/src/sql/orderbook/order_rewards.sql
+++ b/src/sql/orderbook/order_rewards.sql
@@ -45,4 +45,4 @@ from trade_hashes
                        on trade_hashes.order_uid = o.order_uid
                          and trade_hashes.auction_id = o.auction_id
       left outer join winning_quotes wq
-            on o.order_uid = wq.order_uid;
+            on trade_hashes.order_uid = wq.order_uid;


sunce86 (commented):
Is it expected to have `surplus_fee = 0` for market orders?

harisang:
> Is it expected to have `surplus_fee = 0` for market orders?

I have to admit I don't remember on the top of my head how exactly is this table used and where, but indeed up till now, all market orders had a zero entry for `surplus_fee` in the raw_order_rewards table that we were syncing on Dune, which means this was the default behavior.

fhenneke (approved):

sunce86 (approved):

fhenneke:
> Is it expected to have `surplus_fee = 0` for market orders?

At the moment, the surplus fee is only used in one line in the slippage query:
```
atoms_sold - coalesce(surplus_fee, cast(0 as uint256))        as atoms_sold,
```
It is used to correct the amount sold for limit orders to follow the same convention as for market orders: the total amount users send is atoms sold + (fee + surplus fee). It already uses a default value of 0, so uploading 0 is a bit redundant.

In the future we might use this entry to upload the total fee (in sell tokens). But it will not be used for anything except monitoring.

PR #64, opened 2024-01-19
Add protocol fees to dune

fhenneke:
This PR adds protocol fees to dune.

It adds the changes which were also done to the solver rewards repo ([here](https://github.com/cowprotocol/solver-rewards/pull/333)) to the dune sync repo.

The data uploaded to dune changes for `batch_rewards` and `orders_rewards`.

- For batch rewards, we sync an additional entry `protocol_fee` which contains the ETH value of all protocol fees in a batch. Additionally, fees are reduced by network fees. We correct for the fact that the `settlement_observations` table calculates fees from surplus in the sell token while the protocol fee might be taken in the buy token.
- For order rewards, we additionally sync the entries
  - `protocol_fee`: The protocol fee in the token it was charged in.
  - `protocol_fee_token`: The token in which the fee is charged, i.e., sell token for volume based fees and surplus token for surplus based fees. It is `NULL` when there was no protocol fee.
  - `protocol_fee_native_price`: The native price of the protocol fee token in the auction which was settled. It is `0.0` if there is no protocol fee.


src/models/batch_rewards_schema.py
```diff
@@ -32,6 +32,7 @@ def from_pdf_to_dune_records(cls, rewards_df: DataFrame) -> list[dict[str, Any]]
                     "execution_cost": int(row["execution_cost"]),
                     "surplus": int(row["surplus"]),
                     "fee": int(row["fee"]),
+                    "protocol_fee": int(row["protocol_fee"]),
```
harisang, line 35: Should we maybe add this at the very end? Not sure we use indices at any point (I suspect not), but just speculating here...

src/fetch/orderbook.py
```diff
@@ -97,6 +97,7 @@ def get_batch_rewards(cls, block_range: BlockRange) -> DataFrame:
             open_query("orderbook/batch_rewards.sql")
             .replace("{{start_block}}", str(block_range.block_from))
             .replace("{{end_block}}", str(block_range.block_to))
+            .replace("{{EPSILON}}", "10000000000000000")
```
harisang, line 100: Maybe add a comment about what this epsilon is?

src/models/batch_rewards_schema.py
fhenneke: The ordering in this data structure should not matter sice it uses a python dict, which does not have an order.

I think the ordering is a bit inconsistent between different repos. We should probably stick with whatever we find best for readability. We could put it at the end or group with surplus, fees, etc.

src/fetch/orderbook.py
```diff
@@ -97,6 +97,7 @@ def get_batch_rewards(cls, block_range: BlockRange) -> DataFrame:
             open_query("orderbook/batch_rewards.sql")
             .replace("{{start_block}}", str(block_range.block_from))
             .replace("{{end_block}}", str(block_range.block_to))
+            .replace("{{EPSILON}}", "10000000000000000")  # ETH cap for payment (in WEI)
```
harisang, line 100: Not sure if we go that way, but since the new CIP for adjusting and renewing the solver rewards contains an asymmetric cap, we should actually introduce two parameters here.

fhenneke: I mostly introduced the parameter for consistency with the queries in the solver rewards repo.

If we go for an asymmetric cap, we should add that in a separate PR, imo.

src/sql/orderbook/batch_rewards.sql
```diff
@@ -53,32 +180,37 @@ WITH observed_settlements AS (SELECT
...
                                    block_deadline,
                                    solver,
                                    execution_cost,
                                    surplus,
-                                   fee,
-                                   surplus + fee - reference_score           as uncapped_payment_eth,
-                                   -- Uncapped Reward = CLAMP_[-E, E + exec_cost](uncapped_payment_eth)
-                                   LEAST(GREATEST(-10000000000000000, surplus + fee - reference_score),
-                                         10000000000000000 + execution_cost) as capped_payment,
+                                   protocol_fee + fee - network_fee_correction as fee,-- total fee for ranking
```
harisang, line 204: Should we call this something different? I am a bit confused as two lines later you use a `fee` term in the `uncapped_payment_eth` computation, and it is probably not referring to this total fee

fhenneke: We could to `corrected_fee` and later select `corrected_fee as fee` for the uploading. Changing the name for what we upload might be breaking dune queries (but not the queries we use for payments, I guess, which only depend on `capped_payment`).

Another alternative is to repurpose the `fee` field to just be something like network fees. We do not use the field directly in the payment queries, so there it should be fine. Since protocol fees are synced explicitly, we can still recover the corrected quality on dune.

src/sql/orderbook/order_rewards.sql
```diff
@@ -28,21 +128,28 @@ with trade_hashes as (SELECT settlement.solver,
                         WHERE ((o.kind = 'sell' AND o.buy_amount <= oq.buy_amount)
                             OR (o.kind='buy' AND o.sell_amount >= oq.sell_amount))
                           AND o.partially_fillable='f'
-                          AND block_number > {{start_block}}
-                          AND block_number <= {{end_block}}
+                          AND block_number > 19000000
```
harisang, line 131: This was probably accidental I guess

fhenneke: Fixed it.

src/sql/orderbook/batch_rewards.sql
fhenneke: The query now computes `protocol_fee` and `network_fee` separately. The payment is thus by `quality = surplus + protocol_fee + network_fee`.

On dune, the entry `network_fee` is still called just `fee` for now to not break existing queries.

fhenneke:
Comments have been addressed. This is ready to be merged now.

PR #67, opened 2024-01-25
Fix volume cap for protocol fees

fhenneke:
The cap for volume fees in the current code is based on the assumption that the protocol fee on surplus for sell orders is a fraction of the amount received by the user, i.e., a fraction of the total buy amount. The implementation in the driver is based on the cap being a fraction of the amount sent by the user, i.e., a fraction of the total sell amount. This resulted in an issue due to fees in sell tokens being implicitly converted to buy tokens via the uniform clearing price and not using effective directional prices.

This PR changes the cap to be consistent with the current driver implementation.

This requires re-syncing data to dune.

The code should be changed back in combination with a fix to the driver for rank by surplus since the concept of observed fee has no meaning there.


harisang (approved):
Looks good!

PR #69, opened 2024-01-26
Fix volume cap for protocol fee for buy orders

harisang:
Follow up PR on #67 that takes care of volume cap for buy orders.

### DIFF
--- a/src/sql/orderbook/batch_rewards.sql
+++ b/src/sql/orderbook/batch_rewards.sql
@@ -87,30 +87,20 @@ order_surplus AS (
                     CASE
                         WHEN os.kind = 'sell'
                             THEN
-                                CASE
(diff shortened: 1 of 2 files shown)


fhenneke:
I would prefer if the removal of the case for 100% volume cap were done later. Resyncing data for last week might not be possible with the new code.

harisang:
> I would prefer if the removal of the case for 100% volume cap were done later. Resyncing data for last week might not be possible with the new code.

Can you elaborate a bit? Where exactly would the resyncing fail?

fhenneke:
> Can you elaborate a bit? Where exactly would the resyncing fail?

My mistake, it should be fine.

fhenneke (approved):
Ran tests locally and they passed.

Issue #60, opened 2024-01-05
Failing unit test


harisang:
@fhenneke first pointed out that there is some test that is now failing. Since I am personally a bit ignorant about this, any help would be appreciated. One option would be to just remove this test. cc @bh2smith @MartinquaXD 

`=================================== FAILURES ===================================
__________________________ TestIPFS.test_get_content ___________________________

self = <tests.unit.test_ipfs.TestIPFS testMethod=test_get_content>

    def test_get_content(self):
        self.assertEqual(
            {
                "version": "0.1.0",
                "appCode": "CowSwap",
                "metadata": {
                    "referrer": {
                        "version": "0.1.0",
                        "address": "0x424a46612794dbb8000194937834250Dc723fFa5",
                    }
                },
            },
            Cid.old_schema(
                "3d876de8fcd70969349c92d731eeb0482fe8667ceca075592b8785081d630b9a"
            ).get_content(ACCESS_KEY, max_retries=10),
        )
    
>       self.assertEqual(
            {
                "version": "1.0.0",
                "appCode": "CowSwap",
                "metadata": {
                    "referrer": {
                        "kind": "referrer",
                        "referrer": "0x8c35B7eE520277D14af5F6098835A584C337311b",
                        "version": "1.0.0",
                    }
                },
            },
            Cid.old_schema(
                "1FE7C5555B3F9C14FF7C60D90F15F1A5B11A0DA5B1E8AA043582A1B2E1058D0C"
            ).get_content(ACCESS_KEY),
        )
E       AssertionError: {'version': '1.0.0', 'appCode': 'CowSwap'[122 chars]0'}}} != None

tests/unit/test_ipfs.py:73: AssertionError
=========================== short test summary info ============================
FAILED tests/unit/test_ipfs.py::TestIPFS::test_get_content - AssertionError: {'version': '1.0.0', 'appCode': 'CowSwap'[122 chars]0'}}} != None`

bh2smith:
At first looks like an "intermittent failure" possibly because of a failure to retrieve app_data at the given hash. Have you tried running the test locally? Because one would not expect to get None.

harisang:
Will try to do that.

Btw, there is some "inconsistency" in the code, but i have no idea if this affects things. In that test, in some cases we use

`.get_content(ACCESS_KEY, max_retries=10)`

and in some other cases we have  `.get_content(ACCESS_KEY)`

Not sure if this `max_retries` would change anything though

[...]

PR #70, opened 2024-01-26
Remove failing test

harisang:
This PR removes a failing IPFS test, as reported in issue #60 . Although it would be nice to fix, we don't really have the capacity for this now, and so i am proposing to remove this test so that we don't get constant failures in all the PRs that we are doing at the moment.


bh2smith:
You could just put an ignore/skip tag on the test with a todo and corresponding issue to get back to it.

bh2smith:
If this test is consistently failing, this seems to suggest that the only source of truth now is the Dune table. If it were wiped, we might not be able to recreate that record...

fhenneke (approved):

PR #71, opened 2024-01-26
[Hotfix] Update latest block query

fhenneke:
Dune uploads are broken at the moment due to a change to the `settlements` table in our database. This PR changes the query to be compatible with that change.

Instead of looking at the earliest block without some transaction data in the `settlements` table we now look at the latest block in the `settlement_observations` table.

This should work with both the new and old format of the settlements table.

Tests passed locally.

### DIFF
--- a/src/sql/orderbook/latest_block.sql
+++ b/src/sql/orderbook/latest_block.sql
@@ -1,3 +1,2 @@
-select min(block_number) latest
-from settlements
-where tx_from is null;
+select max(block_number) latest
+from settlement_observations;


src/sql/orderbook/latest_block.sql
```diff
@@ -1,3 +1,2 @@
-select min(block_number) latest
-from settlements
-where tx_from is null;
\ No newline at end of file
+select max(block_number) latest
```
harisang, line 1: Should we maybe have
`max(block_number) + 1`
?

Not sure if it makes any sense, but so as to follow the convention used up till now that the latest block is to be ignored (i.e., it is an open inverval)

fhenneke: I am not sure either what makes sense here. From what I can piece together is that the original query returned a block which did not have data in our database and subtracted `65`. This should result in a block number which does have finalized blocks.

With the new query we are (probably?) reading data from a table which only contains information on blocks which are already finalized. This would mean that instead of adding `1` to the largest block number, we could remove the subtraction of `65`.

In any case, the only thing that needs to be prevented is to include blocks in the block range which contain settlements which do not have all the required information in the database yet. So I would opt for not adding `+ 1` and leaving the `- 65`in place. In the future, we should check my argument above and remove the `- 65` if possible.

harisang: Ok, makes sense. I also had a look at the code and saw this `REORG_THRESHOLD = 65` correction, so probably the +1 wouldn't really change much here.

harisang (approved):

PR #76, opened 2024-01-31
Remove auction_transaction table from queries

harisang:
This PR adapts the relevant sql queries that were using the `auction_transaction` table to recover the auction_id, as the table is now deprecated.

### DIFF
--- a/src/sql/orderbook/order_rewards.sql
+++ b/src/sql/orderbook/order_rewards.sql
@@ -6,22 +6,19 @@ with trade_hashes as (SELECT settlement.solver,
                              auction_id
                       FROM trades t
                              LEFT OUTER JOIN LATERAL (
-                        SELECT tx_hash, solver, tx_nonce, tx_from
+                        SELECT tx_hash, solver, tx_nonce, tx_from, auction_id, block_number, log_index
                         FROM settlements s
                         WHERE s.block_number = t.block_number
                           AND s.log_index > t.log_index
                         ORDER BY s.log_index ASC
                         LIMIT 1
                         ) AS settlement ON true
-                             join auction_transaction
-                        -- This join also eliminates overlapping
-                        -- trades & settlements between barn and prod DB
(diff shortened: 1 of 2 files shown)


sunce86 (dismissed):
2 occurrences of `at.auction_id` need to be updated to `s.auction_id`

1. `ON at.auction_id = ss.auction_id`
2. `ON t.order_uid = oe.order_uid AND at.auction_id = oe.auction_id`

src/sql/orderbook/order_rewards.sql
```diff
@@ -110,20 +105,21 @@ order_surplus AS (
...
+    WHERE (o.class = 'market'
+        OR (o.kind = 'sell' AND (oq.sell_amount - oq.gas_amount * oq.gas_price / oq.sell_token_price) * oq.buy_amount >= o.buy_amount *  oq.sell_amount)
+        OR (o.kind='buy' AND o.sell_amount >= oq.sell_amount + oq.gas_amount * oq.gas_price / oq.sell_token_price))
+        AND o.partially_fillable='f' -- the code above might fail for partially fillable orders
+        AND t.block_number > {{start_block}}
+        AND t.block_number <= {{end_block}}
+        AND oq.solver != '\x0000000000000000000000000000000000000000')
 -- Most efficient column order for sorting would be having tx_hash or order_uid first
-select block_number,
+select trade_hashes.block_number as block_number,
```
MartinquaXD, line 122: Does selecting the `block_number` differently change anything?

MartinquaXD (approved):
Adjustments to the query look correct to me.

src/sql/orderbook/order_rewards.sql
harisang: Thought this was leftover redundant change from when i was trying to fix the test, I undid it but then needed to reapply it as indeed there is a complaint about the block_number being ambiguous. it should be the trade_hashes block number though as this is the main table we are left-joining with the rest.

PR #77, opened 2024-02-05
make volume fee code identical to code in solver-rewards repo

fhenneke:
This code change makes the code to compute volume based protocol fees
identical to the code in the solver-rewards repo.

Potentially the implementation is not consistent to what will be done in
the driver. Since volume based fees are not used in production yet this is
not much of an issue for now.

I am also not super sure if we might need different cases for sell and buy orders.
In any case, since we only test those queries in the solver rewards repo
we should make the code consitsant with said repo.

### DIFF
--- a/src/sql/orderbook/batch_rewards.sql
+++ b/src/sql/orderbook/batch_rewards.sql
@@ -97,7 +97,7 @@ order_surplus AS (
                                 )
                     END
             WHEN fp.kind = 'volume'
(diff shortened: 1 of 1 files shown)


harisang:
Approving for consistency but we need to revisit this. There is already an open issue on the solver-rewards repo about this.

harisang (approved):

Issue #72, opened 2024-01-27
Delays when syncing data


harisang:
After PR #71, which fixed the issue with the backend database change, it seems to me that we might sync data with signficant delay, as the syncing, I believe, now depends on both prod and staging having at least one settlement every few hours. The reason for that, I think, is the `min` operator in this line:

https://github.com/cowprotocol/dune-sync/blob/d6f34e042f0426fac0f93d0d9a60c19c9d775df5/src/fetch/orderbook.py#L72

I would propose that we actually change it to `max` now. Let me know what you think @fhenneke

fhenneke:
I would change it to `max`  and remove the `- REORG_THRESHOLD`.

With the new query we look at blocks which have already successfully been processed (since these is an entry in `settlement_observations`). What can then go wrong is that maybe there are staging and production settlements on adjacent blocks and the later settlement is already processed but not the former. Then the former will not end up on dune. Not sure if that is actually happening sometimes. For that purpose we could have some delay, so something like `- MAX_PROCESSING_DELAY` with `MAX_PROCESSING_DELAY = 1` or `2` or whatever seems fit. The old query had a delay of `65` which might included implicitly a `MAX_PROCESSING_DELAY = 1` since the reorg threshold should be `64`.

PR #78, opened 2024-02-05
use more optimistic block range for sync

fhenneke:
This PR addresses issue #72. At the moment the block range is computed using the minimum of the latest finalized block from production and
staging and subtracting `REORG_THRESHOLD = 65`. This was done as a hot fix in #71. One problem with this approach is that having few settlements in staging blocks upload of production data to dune.

This PR changes the behavior to look at the maximum block of the latest finalized blocks from production and staging subtracting
`MAX_PROCESSING_DELAY = 1`. This should leave enough time for the autopilot to process settlements after finalization in staging and production.

For testing we should monitor if there are settlements with missing information on dune compared to our database.

### DIFF
--- a/src/fetch/orderbook.py
+++ b/src/fetch/orderbook.py
@@ -15,7 +15,7 @@ from sqlalchemy.engine import Engine
 from src.models.block_range import BlockRange
 from src.utils import open_query
 
-REORG_THRESHOLD = 65
+MAX_PROCESSING_DELAY = 1
 
 
 class OrderbookEnv(Enum):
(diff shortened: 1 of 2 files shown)


src/fetch/orderbook.py
```diff
@@ -15,7 +15,7 @@
 from src.models.block_range import BlockRange
 from src.utils import open_query
 
-REORG_THRESHOLD = 65
+MAX_PROCESSING_DELAY = 1
```
harisang, line 18:
```suggestion
MAX_PROCESSING_DELAY = 10
```
I would increase the delay a bit just to be on the safe side.

harisang (approved):
Looks good.

PR #79, opened 2024-02-07
Add MATERIALIZED to orderbook queries

fhenneke:
This PR addresses a performance bug in dune-sync orderbook queries.  See https://github.com/cowprotocol/solver-rewards/pull/337 for an explanation.  The problem in dune-sync is not as bad since block ranges are generally smaller than for weekly solver rewards.

The main difference of this PR to https://github.com/cowprotocol/solver-rewards/pull/337 is that the performance bug is present in two queries. Both were sprinkeled with `MATERIALIZED`.

It should be merged only it the other repo is merged. Hopefully it is merged before the next time we need to resync a week of data to dune.


fhenneke:
As this does not seem to be an issue anymore, I will close this.

PR #80, opened 2024-02-08
Implement CIP-36

fhenneke:
This PR implements changes due to CIP-36. See https://github.com/cowprotocol/solver-rewards/pull/339 for more details.

Fewer changes are required in this repository compared to the solver-rewards repo since some of the reward logic is implemented on dune. The respective dune queries have to be adapted as well.

After merging this a resync for the accounting period starting on 06.02.2024 is required.


fhenneke:
I updated the integration test for batch and order rewards to a block range from today.This block range includes an order with protocol fees and capped settlements from above and below, The separate test for protocol fees was removed.

harisang (approved):
looks good!

PR #83, opened 2024-02-12
Add quote information for trades

harisang:
This PR adds raw quote information for each trade.


tests/unit/test_order_rewards_schema.py
```diff
@@ -19,6 +19,14 @@ def test_order_rewards_transformation(self):
                 "protocol_fee": [1000000000000000, 123123123123123, 0],
                 "protocol_fee_token": ["0x91", "0x92", None],
                 "protocol_fee_native_price": [1.0, 0.1, 0.0],
+                "quote_sell_amount": [10000000000000000, 2000000000000000, 35000],
+                "quote_buy_amount": [1000, 2000, 10],
+                "quote_gas_cost": [
+                    5000000000000000,
+                    6000000000000000.52,
+                    12000000000000000.123,
```
fhenneke, line 27: Do we need non-integer gas costs? They should be in Gwei. But due to the tip and base fee mechanism it might be a float, right?

Issue #82, opened 2024-02-09
Error msg about DataFrame concatenation


harisang:
For the past few weeks, we have observed this error showing up in the logs.

`ERROR (pod: dune-sync-order-rewards):
/app/src/fetch/orderbook.py:91: FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.`

Although it doesn't seem to affect anything for now, we should address this soon.

PR #85, opened 2024-03-05
Fix dataframe warning

harisang:
This PR addresses issue #82

### DIFF
--- a/src/fetch/orderbook.py
+++ b/src/fetch/orderbook.py
@@ -90,9 +90,9 @@ class OrderbookFetcher:
         assert set(prod.solver).isdisjoint(set(barn.solver)), "solver overlap!"
         if not prod.empty and not barn.empty:
             return pd.concat([prod, barn])
-        elif not prod.empty:
+        if not prod.empty:
             return prod.copy()
-        elif not barn.empty:
+        if not barn.empty:
             return barn.copy()
         return pd.DataFrame()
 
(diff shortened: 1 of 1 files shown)


src/fetch/orderbook.py
```diff
@@ -88,7 +88,13 @@ def get_order_rewards(cls, block_range: BlockRange) -> DataFrame:
...
         # Solvers do not appear in both environments!
         assert set(prod.solver).isdisjoint(set(barn.solver)), "solver overlap!"
-        return pd.concat([prod, barn])
+        if not prod.empty and not barn.empty:
+            return pd.concat([prod, barn])
+        if not prod.empty:
+            return prod.copy()
+        if not barn.empty:
+            return barn.copy()
+        return pd.DataFrame()
```
fhenneke, line 97: Looks like it does what it is supposed to. The following looks more pythonic, but I have not tested it.
```suggestion
        return pd.concat(df for df in [prod, barn] if not df.empty)
```

fhenneke (approved):
Looks like it does what it is supposed to.

One thing I do not understand is if this fix is equivalent to the old behavior of `concat` or the new behavior? Or does it make no difference in our case?

src/fetch/orderbook.py
harisang: Yeah agreed but i don't feel that comfortable with all these compact pythonic ways, so if you don't mind, i want to test it as is and then we can potentially make it nicer in the future.

Issue #84, opened 2024-02-28
Remove Warehouse db


harisang:
As we want to simplify slippage accounting, we decided, for now, to stop running the internal imbalances job, and shut down the related database. Fixes in the tests in this repo are needed for any future PR to pass all tests.

PR #86, opened 2024-03-05
Comment out warehouse related parts

harisang:
This PR removes/comments out warehouse db references.

We decided to not fully remove those as we might need to revert and restart the internalization simulations.

Addresses issue #84

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -49,4 +49,3 @@ jobs:
         env:
           PROD_DB_URL: ${{ secrets.PROD_DB_URL }}
           BARN_DB_URL: ${{ secrets.BARN_DB_URL }}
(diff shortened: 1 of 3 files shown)


fhenneke (approved):
I do not really know how to check this code. It looks fine, I guess.

In general I would say that commenting out code is less maintainable than removing code. We can always revert this PR if necessary.

If we remove internal imbalances, we should also have a look at the code in solver-rewards and on tenderly.

harisang:
> I do not really know how to check this code. It looks fine, I guess.
> 
> In general I would say that commenting out code is less maintainable than removing code. We can always revert this PR if necessary.
> 
> If we remove internal imbalances, we should also have a look at the code in solver-rewards and on tenderly.

Closing the issue in favor of PR #87 to address reviewer's comment

PR #87, opened 2024-03-05
Remove internal imbalances job

harisang:
This PR removes the code related to the internal imbalances job, as we plan to stop using it.


fhenneke (approved):
Looks good to me. Can have a closer look later (and after merging this) to see if there are is additional dead code somewhere.

PR #88, opened 2024-03-07
[Mirror] Implementation of CIP-38

fhenneke:
This PR mirrors changes to the batch rewards query from https://github.com/cowprotocol/solver-rewards/pull/345.

It should be merged after switching to the new ranking. Data should then be resynced from the block we started with the new ranking.


harisang (approved):
looks good!

PR #91, opened 2024-03-11
Update prod batch/order rewards queries

harisang:
This PR mimics PR #90 in prod. It also introduces separate queries for prod and barn for batch/order rewards, so that we have some flexibility whenever a staging change happens.


fhenneke (approved):
I have not checked the code in detail, but superficially this looks good to me.

PR #94, opened 2024-04-03
Add integration fee recipient

harisang:
This PR adds one more field in the raw_order_rewards table on Dune; namely, what is the recipient of the protocol fee collected, if any, for an order.

Note that here we slightly abuse the term "protocol fee" as integrator fees are not really protocol fees, but from a solvers and backend point of view, they behave identically.


harisang:
Side comment: it ended up being extremely annoying to pass the integration test. Hadn't connected to the database locally so was relying on github for debugging. The block/block_deadline business is quite confusing, so I need to do at least some sanity checks to convince myself everything is good. Opening an issue to address this.

src/models/order_rewards_schema.py
```diff
@@ -35,6 +35,7 @@ def from_pdf_to_dune_records(cls, rewards_df: DataFrame) -> list[dict[str, Any]]
                     "quote_buy_amount": str(row["quote_buy_amount"]),
                     "quote_gas_cost": float(row["quote_gas_cost"]),
                     "quote_sell_token_price": float(row["quote_sell_token_price"]),
+                    "protocol_fee_recipient": row["protocol_fee_recipient"],
```
fhenneke, line 38: I would probably have called it `integrator_fee_recipient` to keep the distinction between protocol fees and integrator fees in the outwards communication. The fact that we internally use the same infrastructure is more of an implementation detail.

PR #99, opened 2024-06-26
Extend Dune Sync appdata on Gnosis Chain

harisang:
This PR will attempt to replicate the mainnet setup in order to sync appdata from Gnosis Chain on Dune. Although wasteful, we need this data soon and as we are not very comfortable/familiar with Dune sync, this seems like the fastest way out.

This is marked as draft as it is incomplete.


src/fetch/dune.py
```diff
@@ -59,27 +59,45 @@ async def fetch(self, query: Query) -> list[DuneRecord]:
...
-        app_hash_query = QUERIES["APP_HASHES"].with_params(
-            block_range.as_query_params()
-        )
+        app_hash_query: Query
+        if chain == "mainnet":
+            app_hash_query = QUERIES["APP_HASHES"].with_params(
+                block_range.as_query_params()
+            )
+        else:
+            if chain == "gnosis":
```
fhenneke, line 98: is this equivalent to an `elif`?

src/fetch/dune.py
```diff
@@ -59,27 +59,45 @@ async def fetch(self, query: Query) -> list[DuneRecord]:
...
+                ]
+            )
         return int(
             # KeyError here means the query has been modified and column no longer exists
             # IndexError means no results were returned from query (which is unlikely).
-            (await self.fetch(QUERIES["LATEST_APP_HASH_BLOCK"].query))[0][
+            (await self.fetch(QUERIES["LATEST_APP_HASH_BLOCK_GNOSIS"].query))[0][
                 "latest_block"
             ]
         )
```
fhenneke, line 84: using an `if .. elif ...` seems clearer here, throwing an exception is the chain name is not correct.
(the pythonic way might be to use a `match ... case ... case ...` block)

src/sync/app_data.py
```diff
@@ -122,26 +122,32 @@ async def fetch_content_and_filter(
...
     config: AppDataSyncConfig,
     ipfs_access_key: str,
     dry_run: bool,
+    chain: str,
 ) -> None:
     """App Data Sync Logic"""
+
+    genesis_bl = 15310317
+    if chain == "mainnet":
+        genesis_bl = 12153262
```
fhenneke, line 137: using an `if .. elif ...` seems clearer here, as above

fhenneke (approved):
Changes look correct but I am not confident at all with that part of the code base.

Is there some way to test changes locally? If not, lets test in our cluster.

PR #102, opened 2024-07-10
Update dune client to latest version

fleupold:
The new version allows for this API (which we will use to rewrite app data fetching): https://docs.dune.com/api-reference/tables/endpoint/upload


bh2smith (approved):
Sorry about the `QueryBase` change, the NPM dune client has a much nicer interface!

https://www.npmjs.com/package/@duneanalytics/client-sdk

PR #103, opened 2024-07-10
Rewrite AppData sync using csv upload feat

fleupold:
This PR dramatically simplifies the way we sync app_data into Dune. Instead of looking for hashes we see on Dune for which we don't have the app_data pre-image and then trying to fetch this data from IPFS, we simply mirror the entire app_data pre-image table we have locally to Dune using their csv upload feature.

We ensured that we are no longer seeing orders for which app_data can only be retrieved from IPFS (to the contrary, we see more and more orders for which the pre image has only been written into the DB).

In order to achieve this, we
- Dramatically reduced the logic inside `sync/app_data.py` and its configuration to basically be a full tablescan from the backend which gets written using the CSV upload feature of the dune python client
- Removed queries that were related to finding missing app data hashes
- Rewrote the e2e test for app data hashes

App data for each network (mainnet, gnosis, arbitrum) will be written in its own table (since each network requires a separate db connection to the designated target db). 
The table name can be specified as an additional parameter (which feels a bit 🤡, I'm not sure how sync job specific arguments were envisioned in the current architecture)

### Test Plan

Both `python3 -m src.main --sync-table app_data` as well as `python -m pytest tests/e2e/test_sync_app_data.py` pass


bh2smith (approved):
So glad you are finally using the upload CSV endpoint! I mentioned this to @fhenneke several months ago. Note performing this regularly will consume API credits (while the old solution would have remained free indefinitely). 

Note that you can now also remove all the AWS dependencies and utilities from this code base (to make sure you caught everything).

src/sync/app_data.py
```diff
@@ -1,154 +1,27 @@
...
-        ipfs_access_key=ipfs_access_key,
-        missing_file_name=config.missing_files_name,
-    )
-    await data_handler.fetch_content_and_filter(
-        max_retries=config.max_retries, give_up_threshold=config.give_up_threshold
-    )
-    UploadHandler(aws, data_handler, table=SYNC_TABLE).write_and_upload_content(dry_run)
+    hashes = orderbook.get_app_hashes()
+    if not dry_run:
+        dune.upload_csv(
```
harisang, line 21: Just to double-check, this essentially kills the old table and creates a new one with the updated csv, right?

tests/e2e/test_sync_app_data.py
```diff
@@ -1,80 +1,50 @@
...
     def setUp(self) -> None:
         load_dotenv()
-        self.dune = DuneFetcher(os.environ["DUNE_API_KEY"])
-        self.config = AppDataSyncConfig(
-            volume_path=Path(os.environ["VOLUME_PATH"]),
-            missing_files_name="missing_app_hashes.json",
-            max_retries=2,
-            give_up_threshold=3,
+        self.dune = DuneClient(os.environ["DUNE_API_KEY"])
+        self.fetcher = DuneFetcher(self.dune)
```
fhenneke, line 20: `DuneFetcher` does not seem to be used anywhere. Would it be safe to remove that class?

src/main.py
```diff
@@ -50,36 +50,32 @@ def __init__(self) -> None:
     volume_path = Path(os.environ["VOLUME_PATH"])
     args = ScriptArgs()
     aws = AWSClient.new_from_environment()
+    dune = DuneClient(os.environ["DUNE_API_KEY"])
+    orderbook = OrderbookFetcher()
 
     if args.sync_table == SyncTable.APP_DATA:
+        table = os.environ["APP_DATA_TARGET_TABLE"]
+        assert table, "APP_DATA sync needs a DUNE_NETWORK_NAME env"
```
fhenneke, line 58: update string to use `APP_DATA_TARGET_TABLE`

fhenneke (commented):
This is a great change.

Potentially the whole `DuneFetcher` code can be removed as well.

We should eventually move other tables to the new API as well.

fleupold:
>  Be aware that this endpoint consumes API credits, so performing this regularly might use a lot

I wasn't able to find any details on how much credits are used (so I ended up thinking it might be free). @bh2smith do you know where this is documented? Not [here](https://docs.dune.com/api-reference/tables/endpoint/upload) afaict.

bh2smith:
Maybe it's just part of the storage plan on the data set.

src/sync/app_data.py
fleupold: Yes, it overrides any table with the same name

PR #104, opened 2024-07-11
[Mirror] Use scores instead of surplus for reward computation

fhenneke:
This PR mirrors changes from https://github.com/cowprotocol/solver-rewards/pull/372

The changes in the test are due to rounding errors in computing scores in the driver and in the old SQL query.


harisang (approved):
Identical to solver rewards PR that i reviewed and tested

PR #105, opened 2024-07-15
[Trivial] remove optional env vars from critical path

fleupold:
VOLUME_PATH and AWS env vars aren't needed anymore for the app_data job and so they shouldn't be required in the environment (cf. https://github.com/cowprotocol/infrastructure/pull/1770/files#diff-6ed79a81a644cd7d10f8cdd1395db1c2371d171337c95f5699db5c906a0b8682R44-R48)

### Test Plan
Run `python3 -m src.main --sync-table app_data` without any of the unrelated env vars defined.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -47,9 +47,7 @@ class ScriptArgs:
 
 if __name__ == "__main__":
     load_dotenv()
-    volume_path = Path(os.environ["VOLUME_PATH"])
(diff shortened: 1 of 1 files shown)


harisang (approved):
Makes sense

bh2smith (approved):

PR #106, opened 2024-07-30
Fix for week where Copium Capital used the same account for both staging and prod

harisang:
The script currently produces an error if the same address was used both for staging and prod submissions. This was indeed the case for Copium Capital that tested colocation in staging before we enabled them with the same account in prod (and disabled them in staging at the same time).

The proposed solution is really to bypass the check just for this one occasion, by pretending that the staging submissions of Copium Capital didn't happen from address `0x008300082C3000009e63680088f8c7f4D3ff2E87`,  but instead from the old staging address that the core team controls.


fhenneke (approved):
Looks reasonable

PR #107, opened 2024-07-31
revert copium capital specific changes

harisang:
Pr #106 was a temporary workaround that is not needed anymore (hoping here we won't need to resync data from these weeks).


fhenneke (approved):
Next time you could actually just [revert the commit](https://git-scm.com/docs/git-revert).

PR #108, opened 2024-08-14
Mirror changes from solver rewards

fhenneke:
This PR mirrors changes to solver rewards from #372 and #381.

Tests were adapted since the computation of rewards was changed from recomputing the winning score to using what the driver reports.


bram-vdberg (approved):
LGTM!

PR #114, opened 2024-10-11
Simplify protocol fee calculations and remove consistency rewards (mirroring solver rewards changes)

harisang:
This PR mirrors the changes in the solver rewards repo from the following two PRs:
- https://github.com/cowprotocol/solver-rewards/pull/401
- https://github.com/cowprotocol/solver-rewards/pull/404

Concretely, the calculation of protocol fees in the sql query is simplified by a lot, and consistency rewards are removed as they are no longer used.

**Note**: some tests needed updating since the new way of fetching protocol fees is not available for the block range the test was set up for. The tests were updated accordingly, and one (bit strange) way to verify the numbers is to run the following dune query, as the data for that block range on Dune have been uploaded via the old (and NOT the proposed) and tested way. The tiny discrepancies are a result of rounding "errors", but they are very small.

For the batch rewards test, this query should be used

```
select
    block_number,
    block_deadline,
    tx_hash,
    solver,
    data.execution_cost,
    data.surplus,
    data.protocol_fee,
    data.fee,
    data.uncapped_payment_eth,
    data.capped_payment,
    data.winning_score,
    data.reference_score
from cowswap.raw_batch_rewards where block_deadline >= 20936269 and block_deadline <= 20936280
```

while for the order rewards test, this query should be used

```
select
    block_number,
    order_uid,
    solver,
    data.quote_solver,
    tx_hash,
    data.surplus_fee,
    data.amount,
    data.protocol_fee,
    data.protocol_fee_token,
    data.protocol_fee_native_price,
    data.quote_sell_amount,
    data.quote_buy_amount,
    data.quote_gas_cost,
    data.quote_sell_token_price,
    data.partner_fee,
    data.partner_fee_recipient,
    data.protocol_fee_kind
from cowswap.raw_order_rewards where block_number>= 20934809 and block_number <= 20934808 + 17 order by block_number ASC
```

As for the integrator fees test, this range on Dune should be user: `block_number>= 20934805 and block_number <= 20934809`


src/sql/orderbook/prod_order_rewards.sql
```diff
@@ -33,293 +33,148 @@ with trade_hashes as (
...
-            WHEN kind = 'sell' then buy_amount + protocol_fee_first
-            ELSE buy_amount
-        END as buy_amount,
-        surplus + protocol_fee_first as surplus,
-        price_improvement + protocol_fee_first as price_improvement,
-        protocol_fee_kind_first,
-        protocol_fee_first,
-        partner_fee_recipient
-    from order_protocol_fee_first
+        trade_data_unprocessed tdu LEFT OUTER JOIN protocol_fee_kind pfk ON tdu.order_uid = pfk.order_uid AND tdu.auction_id = pfk.auction_id
```
fhenneke, line 132: To make it consistent with the other formating:
```suggestion
        trade_data_unprocessed tdu
            LEFT OUTER JOIN protocol_fee_kind pfk
            ON tdu.order_uid = pfk.order_uid AND tdu.auction_id = pfk.auction_id 
```

src/sql/orderbook/prod_order_rewards.sql
```diff
@@ -33,293 +33,148 @@ with trade_hashes as (
...
+    FROM fee_policies fp JOIN trade_hashes th ON fp.auction_id = th.auction_id AND fp.order_uid = th.order_uid
+    GROUP BY fp.auction_id, fp.order_uid
+),
+protocol_fee_kind AS (
+    SELECT
+        fp.auction_id,
+        fp.order_uid,
+        fp.kind
+    FROM fee_policies fp JOIN protocol_fee_policies_application_order pfpao ON fp.auction_id = pfpao.auction_id AND fp.order_uid = pfpao.order_uid AND fp.application_order = pfpao.application_order
+),
```
fhenneke, line 72: I am a bit uneasy about introducing changes here which are not part of the solver-rewards query. But since we plan to revamp some of them, this should be fine for some time.

An alternative implementation would be
```suggestion
protocol_fee_kind AS (
    SELECT DISTINCT ON (fp.auction_id, fp.order_uid)
        fp.auction_id,
        fp.order_uid,
        fp.kind
    FROM fee_policies fp
        JOIN trade_hashes th
        ON fp.auction_id = th.auction_id AND fp.order_uid = th.order_uid
    ORDER BY fp.auction_id, fp.order_uid, fp.application_order
),
```

fhenneke (commented):
Generally looks good.

The only thing missing are new tests for order rewards, right?

src/sql/orderbook/prod_order_rewards.sql
harisang: Note that this part actually changes also the semantics of the protocol_fee_kind entry on Dune, which now always refers to the protocol fee kind and NOT the partner fee kind associated with the order.

I guess something a bit more explicit that would be easier for me to read would be to add the `limit 1` suffix. But what you suggest also is fine.

harisang:
> Generally looks good.
> 
> The only thing missing are new tests for order rewards, right?

This is now addressed. It would be good to check old vs new behavior (by recovering the results from Dune). It seems there are some discrepancies, which i believe should be due to rounding/slightly different ways of computing protocol fees, but would like to check 1-2 examples in more detail.

fhenneke (approved):

harisang:
I ran the proposed order rewards query for the accounting period of Oct 8-15, 2024, and it seems that indeed results look fine.

https://dune.com/queries/4186410

The only non-trivial discrepancies where showing up for trades involving this token: https://etherscan.io/token/0x30ae41d5f9988d359c733232c6c693c0e645c77e

which is a token with 0 decimals. Not sure how we handle such tokens throughout our scripts tbh.

So i am fine merging this PR, if you also agree @fhenneke

PR #115, opened 2024-10-15
Prefect

bram-vdberg:
This PR creates a Prefect deployment to run the dune-sync order rewards job once every 3 hours.


Dockerfile.prefect
```diff
@@ -0,0 +1,14 @@
+FROM python:3.10-slim
+
+WORKDIR /deployments
+
+COPY . .
+
+RUN apt-get update \
+    && apt-get -y install libpq-dev gcc \
+    && pip install psycopg2
```
fhenneke, line 9: Why is this required?

.gitignore
```diff
@@ -129,3 +129,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Vim swap files
+**/*.swp
```
fhenneke, line 134: This is fine, I guess, but my personal preference would be to have a small `.gitignore` with project specific content. For local setup, I use `.git/info/exclude`.

Is this approach with a huge .gitignore better for, say, using dev containers?

requirements/prefect.txt
```diff
@@ -0,0 +1,4 @@
+prefect==3.0.2
```
fhenneke, line 1: Alternatively, we can start with a `prefect.in` and generated `prefect.txt` right away here.

We could also make that a separate change, but I do not see much momentum for any type of refactoring.

requirements/prefect.txt
```diff
@@ -0,0 +1,4 @@
...
+psycopg2==2.9.9
+dune_client==1.7.7
+requests==2.32.3
```
fhenneke, line 4: prefect_github is probably missing somewhere

prefect/deployment.py
```diff
@@ -0,0 +1,144 @@
...
+    table_name = upload_data_to_dune(data, blockrange.block_from, blockrange.block_to)
+    update_aggregate_query(table_name)
+
+
+if __name__ == "__main__":
+    github_repository_block = GitHubRepository.load("dune-sync")
+    deployment = Deployment.build_from_flow(
+        flow=order_rewards,
+        name="dune-sync-prod-order-rewards",
+        schedule=(CronSchedule(cron="0 */3 * * *")), # Once every 3 hours
```
fhenneke, line 138: missing import?

prefect/deployment.py
```diff
@@ -0,0 +1,144 @@
...
+        storage=github_block,
```
fhenneke, line 139: `github_repository_block`?

fhenneke (commented):
I have not checked the actual code in detail and was not yet able to run it locally. What are the required steps?

bram-vdberg:
> I have not checked the actual code in detail and was not yet able to run it locally. What are the required steps?

Running it locally requires changing the way we serve the deployment. I'll make a local_deploy.py file to make this easier so you can just run `docker compose up` instead.

prefect/deployment.py
```diff
@@ -0,0 +1,144 @@
...
+        logger.info(f"Table already in query, not updating query")
+
+
+@flow(retries=3, retry_delay_seconds=60, log_prints=True)
+def order_rewards():
+    logger = get_run_logger()
+    blockrange = get_block_range()
+    orderbook = fetch_orderbook(blockrange)
+    data = cast_orderbook_to_dune_string(orderbook)
+    table_name = upload_data_to_dune(data, blockrange.block_from, blockrange.block_to)
```
fhenneke, line 129: One could also use the block range object here. Would look cleaner here but less clean in `upload_data_to_dune`. Since the latter only uses it for creating a string, it should be fine to look less clean there.

fhenneke (commented):
I like the structure of the `order_rewards` flow. The logic is really clear there.

I do not fully understand the logic behing the batching in time. Some clarifications would help there. 

There are a few unused variables here and there.

Otherwise it looks good for finally using prefect systematically.

.gitignore
bh2smith: Or a global .gitignore

In `~/.gitconfig`:

```
[core]
	excludesfile = /Users/bh2smith/.gitignore_global
```

which contains things like:

```
TODO.md
.venv/
.env.bkp
env/
venv/
.DS_Store
.idea/
.vscode/
out/

```

bram-vdberg:
> I have not checked the actual code in detail and was not yet able to run it locally. What are the required steps?

I made some updates, now you should be able to run the prefect server with `make prefect`. Then you can find the dashboard on `localhost:4200`. There's a local config in `src.prefect.local_deploy` that deploys the flow to this local dashboard. You can test it with: `make deployment`.

One note on running the flow locally: the dune update query command requires an enterprise API key, so I don't use that step on the local deploy. It will just run the commands and upload the data to your dune account.

Dockerfile.prefect
bram-vdberg: I don't use the Dockerfile anymore so it's not needed  anymore. These are needed if you're using sqlalchemy inside a Docker container.

.gitignore
bram-vdberg: > Is this approach with a huge .gitignore better for, say, using dev containers?

Not that I know of, I didn't know you could add it to .git/info/exclude. Thanks!

requirements/prefect.txt
bram-vdberg: Thanks! I added it

prefect/deployment.py
```diff
@@ -0,0 +1,144 @@
...
+    current_weekday = now.weekday()
+    days_since_last_monday = current_weekday if current_weekday != 0 else 7
+    last_monday = now - timedelta(days=days_since_last_monday)
+    last_monday_midnight = last_monday.replace(hour=0, minute=0, second=0, microsecond=0)
+    timestamp = int(last_monday_midnight.timestamp())
+    return timestamp
+
+
+@task
+def get_block_range() -> BlockRange:
```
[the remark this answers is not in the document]
bram-vdberg: At the moment it executes the job once every 30 minutes, but it's very easy to change it. I think Haris recommended 3 hours so I can set it to that. Every time this job runs it re-uploads the entire data for that week, overwriting the current table (for that week). When we get to another week the start block will change and it will instead upload to a new table with a new name. 

So on a transition between weeks it will run the query multiple times for that day, and re-upload each time, changing the table when the startblock moves up a week.

fhenneke (approved):
Approving this so we can test the github flow.

Local testing kind of worked but there are still a few issues. Creating a PR for that would be good. (e.g. dependency conflicts).

PR #117, opened 2024-10-17
Fix prefect workflow

bram-vdberg:
This PR fixes the prefect worker.


harisang:
Not sure i can provide any meaningful review here. Approving so that we move forward with things

harisang (approved):

PR #120, opened 2024-10-25
Remove special handling of empty dataframes

fhenneke:
This PR simplifies how dataframes are concatenated after fetching from prod and barn. The behavior should be identical, except for the unnecessary copies.

The change was introduces in #85 to avoid a deprecation warning. This is not required anymore.

### DIFF
--- a/src/fetch/orderbook.py
+++ b/src/fetch/orderbook.py
@@ -12,8 +12,11 @@ from pandas import DataFrame
 from sqlalchemy import create_engine
 from sqlalchemy.engine import Engine
 
+from src.logger import set_log
 from src.models.block_range import BlockRange
 from src.utils import open_query
+
+log = set_log(__name__)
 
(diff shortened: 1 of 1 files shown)


bram-vdberg (approved):

bram-vdberg:
>The change was introduces in https://github.com/cowprotocol/dune-sync/pull/85 to avoid a deprecation warning. This is not required anymore.
Why is it not required anymore? If there was a version change it might be good to document that in this PR as well.

fhenneke:
The change is not correct as it would still lead to a `FutureWarning` if one of the data frames is empty. The warning is about a change of behavior in determining data types when one dataframe is empty but has a conflicting datatype with the other data frame. This seems to happen here.

A full fix would require also changing the creation of the dataframe to use no types or compatible types.

Closing this for now as there are more important changes to work on.

PR #121, opened 2024-10-26
Remove prefect from deployment

harisang:
I believe the deployment is a bit messed up, as we currently have dune-sync only deployed in prod while prefect lives in staging. The plan is to migrate everything to staging, but until then, any new release issued in prod cannot work; i see these errors
`
stream logs failed container "batch-rewards" in pod "dune-sync-mainnet-batch-rewards-cj-..." is waiting to start: trying and failing to pull image for analytics/dune-sync-mainnet-batch-rewards-cj-...(batch-rewards) `

and i believe the reason is prefect. This PR removes the prefect deployment until we sort things out.

UPDATE: PR was applied and it worked but the latest run revealed another bug; PR #114 had not done a proper cleanup of the references to participation. Given it's a Saturday night and the script is still crashing, i decided to push the fix for that bug in this PR, which basically now serves as the working version of what I am tempting to deploy. We can discuss how to proceed with these on Monday

PR #123, opened 2024-10-30
Remove any leftover references of participation rewards

harisang:
This PR cleans up some leftover references to participation rewards


fhenneke (approved):
LGTM.

Do you know why the code currently is not crashing? The query does not return participation rewards, right?

harisang:
> LGTM.
> 
> Do you know why the code currently is not crashing? The query does not return participation rewards, right?

Yeah, the PR here was replicated in Pr #121, which was then applied as a hotfix in prod (and is still currently running). Will issue a new release now with this PR

PR #124, opened 2024-10-30
Split prefect deployments into dev and prod

bram-vdberg:


harisang:
Should we rename some files? I see names such as `dev_deployment` that seem to only be about order rewards. Or is it that we will add each new jobs to all these files?

fleupold (commented):
I find the use of prefect in this public repository very weird. Imo prefect is part of our infrastructure so all logic around which workflows to run and how often should be configured in the infra repo. Dune-sync can trigger a build or release which can then cause some pipeline in the infra repo to run, but I don't think we should add all this stuff here.

Are we also planning to add a bunch of prefect code to token imbalances and the other data scripts that we need to run periodically?

bram-vdberg:
Closing to move Prefect configs to a separate repository.

PR #125, opened 2024-11-04
Sync price feed

harisang:
This PR adds a simple job that pushes all prices collected in the `prices` table of the analytics db to Dune. It mimics PR #103, and is meant to run as a cronjob. I followed the "legacy method" as there are some concerns about the current way Prefect has been integrated in this repo.

I also prepared this PR in the infra repo. https://github.com/cowprotocol/infrastructure/pull/2217

Note: there are currently not any tests. I saw there are some tests for the app-data sync, so i could try to basically copy-paste those, as i am not sure what's the best way to test this

### DIFF
--- a/src/fetch/orderbook.py
+++ b/src/fetch/orderbook.py
@@ -28,6 +28,7 @@ class OrderbookEnv(Enum):
 
     BARN = "BARN"
     PROD = "PROD"
+    ANALYTICS = "ANALYTICS"
 
     def __str__(self) -> str:
         return str(self.value)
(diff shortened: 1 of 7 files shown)


fhenneke (approved):
Code looks reasonable.

If you do not intend to add tests in this PR, please add an issue and add tests later. We _need_ to test stuff like this.

harisang:
> Code looks reasonable.
> 
> If you do not intend to add tests in this PR, please add an issue and add tests later. We _need_ to test stuff like this.

How would a test look like though?

As a sanity check, one can set
```
PRICE_FEED_TARGET_TABLE = "price_feed_ethereum"
DUNE_API_REQUEST_TIMEOUT = 600
```

run

` python -m src.main --sync-table price_feed`

and confirm the table on dune is being updated.

But i am not sure what a test would check, except for basically having rerun the above command and syncing into some test table.

PR #129, opened 2024-11-06
Setup pip

bram-vdberg:


setup.py
```diff
@@ -0,0 +1,12 @@
+from setuptools import setup
+
+setup(
+    name="dune-sync",
+    version="0.1",
```
fhenneke, line 5: Is this a dummy version which is not to be used? Otherwise, we could use the actual version number from the latest release.

setup.py
fhenneke, line 1: Did you also look into using a pyproject.toml? Is sounds like a more recent standard but might not fit to what you want to achieve.

fhenneke (approved):
I checked that installing works. Not sure how to test if what is installed actually works.

setup.py
bram-vdberg: We could do that and I think it is the better approach, but I don't think we need to. The idea was just to add a simple way to install the repository with PIP. We can consider using a pyproject.toml later if we want to add more configurations.

I can add a line to the README so we can document the reasoning behind this decision for future reference if you want.

fhenneke:
I checked that installing works. How can I check that the installed package is working as expected? Running `from dune_sync import main` gives the error `ModuleNotFoundError: No module named 'dotenv'`. As if dependencies have to installed separately.

bram-vdberg:
> I checked that installing works. How can I check that the installed package is working as expected? Running `from dune_sync import main` gives the error `ModuleNotFoundError: No module named 'dotenv'`. As if dependencies have to installed separately.

I set it up without the requirements, assuming that the person doing the pip install would do that. I updated it to automatically install the prod requirements.txt as dependencies when you install it with PIP.

harisang:
Shouldn't the readme file be updated as well here?

bram-vdberg:
> Shouldn't the readme file be updated as well here?

Thanks, I updated the readme

harisang (approved):

PR #130, opened 2024-11-08
Add config file for multiple envs and chains

bram-vdberg:
This PR adds a config file for running the dune-sync script in multiple environments and chains.

### DIFF
--- a/src/models/config.py
+++ b/src/models/config.py
@@ -1,0 +1,100 @@
+"""Dataclasses for the prefect deployments"""
+import os
+from enum import Enum
+from dataclasses import dataclass, field
+
+from dotenv import load_dotenv
+
+load_dotenv()
+
+
+class ENV(Enum):
+    """
+    Enum ENV class to change environment variables for DEV And PROD
+    """
+
+    DEV = "DEV"
+    PR = "PR"
+    PROD = "PROD"
+
+    def is_dev(self) -> bool:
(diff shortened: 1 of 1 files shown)


src/models/config.py
```diff
@@ -0,0 +1,100 @@
+"""Dataclasses for the prefect deployments"""
```
fhenneke, line 1: Is this only related to prefect? Then it would probably belong into that repo.

bram-vdberg: Sorry, I forgot to change the comment at the top. I moved this to the dune-sync repo since I thought it would be more suitable to hold chain/env specific logic here. But looking at the code again I think you're right. If we refactor this to be chain and env agnostic then these configs won't have much logic anyway.

Do you think we'll be setting specific logic in the dune-sync repo based on env/chain?

fhenneke: If I understood correctly, starting with dune-sync v2, we will not have any cow protocol related code in that repo. All config related to our infrastructure (e.g. prod/barn, which queries to use, how infra changes are tested) will not be part of dune-sync.

You could also have a look at https://github.com/bh2smith/dune-sync to see what the current plan. Providing feedback over there if something is missing might also be useful.

bram-vdberg: Thanks, I'll close this PR and also look at that repo.

PR #132, opened 2024-11-12
Update sqlalchemy to 2.0 and update setup.py

bram-vdberg:
This PR updates the sqlalchemy version of the dune-sync repo to <3.0. This is needed for compatibility with Prefect. It also updates the setup.py to include the .sql files.

### DIFF
--- a/requirements/prod.txt
+++ b/requirements/prod.txt
@@ -6,4 +6,4 @@ pandas==2.1.4
 ndjson>=0.3.1
 py-multiformats-cid>=0.4.4
 boto3>=1.26.12
-SQLAlchemy<2.0
+SQLAlchemy<3.0
--- a/requirements/prod.txt
+++ b/requirements/prod.txt
@@ -6,4 +6,4 @@ pandas==2.1.4
 ndjson>=0.3.1
 py-multiformats-cid>=0.4.4
 boto3>=1.26.12
-SQLAlchemy<3.0
+SQLAlchemy>=2.0,<3.0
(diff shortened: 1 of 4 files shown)


requirements/prod.txt
```diff
@@ -6,4 +6,4 @@ pandas==2.1.4
 ndjson>=0.3.1
 py-multiformats-cid>=0.4.4
 boto3>=1.26.12
-SQLAlchemy<2.0
+SQLAlchemy<3.0
```
bh2smith: I think you should be more explicit here and use >=2.0 since that what it seems like you're aiming for.

bh2smith (commented):
Might not hurt to introduce a lock file to this project.

bh2smith (approved):
So there were like zero changes to be implemented with the jump from <2 to >=2?

Have we considering trying to take one last leap to v3 if its really that easy?

Anyway, this is great as is for its purpose.

bram-vdberg:
> So there were like zero changes to be implemented with the jump from <2 to >=2?
> 
> Have we considering trying to take one last leap to v3 if its really that easy?
> 
> Anyway, this is great as is for its purpose.

Apparently, initially I [made some changes](https://github.com/cowprotocol/dune-sync/pull/132/commits/e546d8641930ef8e97e2bcacc53d29b7c307d184) but it still seems to run properly without them and the unit+integration tests pass, so I don't think it's needed. It might be good if someone with more knowledge of the repo double checks just to be sure.

The Prefect infrastructure we're running this on requires `sqlalchemy[asyncio] >= 2.0, < 3.0.0` so upgrading to 3.0 would make it incompatible with that.

fhenneke (approved):

PR #138, opened 2024-11-15
Add optional env variables to bypass runtime arguments

harisang:
This PR introduces two environment variables, named `arg` and `table`, that are only meant to be used from within the Prefect infrastructure to set up relevant jobs.

An obvious problem is that this leaks implementation details and/or restrictions of Prefect into this repo.

Could an alternative be to actually pass parameters to the main function? Not sure if that is very bad practice or not.


bram-vdberg:
Closing the PR since this change is no longer needed.

PR #140, opened 2024-11-21
Introduce new logic for syncing raw batch data

harisang:
This PR introduces an alternative way of syncing raw batch data on Dune, by relying on the dune api for uploading data as csv files.

The main ideas:
- It creates one table per month
- in order to ensure all data is uploaded, on the first day of each month, it recomputes the previous month's table and re-uploads it
- it finally introduces the auction id and environment as entries in the` batch_data` table!

To test it, one can run

`python -m src.main --sync-table batch_data `

and check the results by running the following query on Dune:

`select * from dune.cowprotocol.dataset_test_batch_rewards_ethereum_2024_11 order by block_deadline DESC`


src/fetch/orderbook.py
```diff
@@ -166,6 +166,82 @@ def get_batch_rewards(cls, block_range: BlockRange) -> DataFrame:
...
+                "{{EPSILON_UPPER}}", "12000000000000000"
+            )  # upper ETH cap for payment (in WEI)
+            .replace("{{env}}", "prod")
+        )
+        batch_data_query_barn = (
+            open_query("orderbook/batch_data.sql")
+            .replace("{{start_block}}", str(block_range.block_from))
+            .replace("{{end_block}}", str(block_range.block_to))
+            .replace(
+                "{{EPSILON_LOWER}}", "10000000000000000"
```
fhenneke: The value of the cap needs to be adapted for Gnosis and Arbitrum One

src/fetch/orderbook.py
```diff
@@ -166,6 +166,82 @@ def get_batch_rewards(cls, block_range: BlockRange) -> DataFrame:
...
+    @classmethod
+    def get_batch_data(cls, block_range: BlockRange) -> DataFrame:
+        """
+        Decomposes the block range into buckets of 10k blocks each,
+        so as to ensure the batch data query runs fast enough.
+        At the end, it concatenates everything into one data frame
+        """
+        start = block_range.block_from
+        end = block_range.block_to
+        bucket_size = 10000
```
fhenneke: This is a constant and could be moved to the front of the file.

src/main.py
```diff
@@ -98,6 +106,18 @@ def main() -> None:
             fetcher=orderbook,
             dry_run=args.dry_run,
         )
+    elif args.sync_table == SyncTable.BATCH_DATA:
+        table = os.environ["BATCH_DATA_TARGET_TABLE"]
```
fhenneke, line 115: This should be added to the `.env.sample` file.

src/sync/batch_data.py
```diff
@@ -0,0 +1,46 @@
...
+
+async def sync_batch_data(
+    node: web3,
+    orderbook: OrderbookFetcher,
+    dune: DuneClient,
+    config: BatchDataSyncConfig,
+    dry_run: bool,
+) -> None:
+    """Batch data Sync Logic"""
+    network = os.environ["NETWORK"]
```
fhenneke, line 26: This should be added to `.env.sample`

src/main.py
```diff
@@ -57,6 +64,7 @@ def main() -> None:
         request_timeout=float(os.environ.get("DUNE_API_REQUEST_TIMEOUT", 10)),
     )
     orderbook = OrderbookFetcher()
+    web3 = Web3(Web3.HTTPProvider(os.environ.get("NODE_URL")))
```
fhenneke: `.env.sample`

fhenneke (commented):
Changes look reasonable.

What is the strategy with respect to rerunning old month? It looks as if the script is not invoced with a parameter and always just uses the current time to determine the month.

harisang:
> Changes look reasonable.
> 
> What is the strategy with respect to rerunning old month? It looks as if the script is not invoced with a parameter and always just uses the current time to determine the month.

So currently it's a very simple strategy. Basically if it is the first day of the month, just rerun the last month as well. Here it assumes that the script will always run at least once every 24h. Of course, this ideally should be fixed, and i would like to add a runtime argument to specify the starting month as well, in case we want to run it again for several months

src/fetch/orderbook.py
harisang: Oh very good point!

src/fetch/orderbook.py
```diff
@@ -45,7 +46,14 @@ class OrderbookFetcher:
...
         load_dotenv()
-        db_url = os.environ[f"{db_env}_DB_URL"]
+        if "NETWORK" in os.environ:
+            db_url = (
+                os.environ[f"{db_env}_DB_URL"]
+                + "/"
+                + os.environ.get("NETWORK", "mainnet")
+            )
+        else:
+            db_url = os.environ[f"{db_env}_DB_URL"]
```
harisang, line 56: This PR changes the meaning of the `PROD_DB_URL` and `BARN_DB_URL` env variables, and this `else` statement just ensures that it defaults to the "old" way, so that the tests won't fail. This should be removed once the secrets in the repo are edited.

